PR #27, opened 2022-01-14
deps: Update Rust to 1.58.

milesj:
Fixes some new lint rules.


.rustrc
```diff
@@ -0,0 +1 @@
+1.58.0
```
milesj, line 1: Not a real thing, but nice to convey to others.

PR #29, opened 2022-01-28
new: Add a dependency graph with orchestration.

milesj:
Converts the `TaskGraph` into a new `DepGraph` that is a DAG under the hood. This allows us to run tasks topologically using the orchestrator/thread pool.


crates/workspace/src/orchestrator.rs
```diff
@@ -0,0 +1,60 @@
+use crate::dep_graph::{DepGraph, Node};
```
milesj, line 1: Need to make stuff `Sync` safe before finishing.

crates/workspace/src/workspace.rs
```diff
@@ -190,8 +186,12 @@ impl Workspace {
             root: root_dir,
             toolchain,
             tsconfig_json,
-            vcs,
             working_dir,
         })
     }
+
+    /// Detect the version control system currently being used.
```
milesj, line 193: Moving to a func avoids having to make it sync safe.

PR #40, opened 2022-02-07
new: Make everything async/await.

milesj:


milesj:
Broke this up into separate PRs to fix the async recursion problem.

PR #77, opened 2022-03-28
Changing URL's to automatic URL's to suppress warnings

malkir:
Building documentation resulted in warnings due to bare URL's, switching to automatic URL's to resolve


milesj:
@malkir These comments are actually used by JsonSchema, so they shouldn't change.

malkir:
Told linter to ignore these warnings just for these files.

PR #129, opened 2022-06-04
[nit] Removing extra whitespace from template

jpoz:
The current templates leaves extra whitespace.


milesj:
@jpoz Do you have an example before/after?

milesj:
Fully testing these here: https://github.com/moonrepo/moon/pull/145

PR #130, opened 2022-06-04
new: Support globs for `.moon/workspace.yml` `projects`.

milesj:
Right now, `projects` support a map of name to source. This works really well, but is also a bit tedious to manage. For open source projects, it's actually more of a burden.

So this PR updates the `projects` setting to also support an array of globs, but this has a few caveats:

- Project IDs use the name of the folder, and are cleaned to our valid list of characters. These IDs run the risk of collision, so instead of overwriting an existing one, we skip the new one and log a warning (this is the same code as `init`).
- Globbing can be a heavy process, especially if there are hundreds of projects. So we're also going to cache the results of the glob with a small lifetime to avoid the glob cost.

### DIFF
--- a/crates/cache/Cargo.toml
+++ b/crates/cache/Cargo.toml
@@ -7,10 +7,8 @@ edition = "2021"
 moon_config = { path = "../config"}
 moon_error = { path = "../error"}
 moon_logger = { path = "../logger"}
-moon_project = { path = "../project"}
 moon_utils = { path = "../utils"}
 serde = { version = "1.0.137", features = ["derive"] }
-sha2 = "0.10.2"
 
 [dev-dependencies]
 assert_fs = "1.0.7"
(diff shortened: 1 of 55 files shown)


crates/cache/Cargo.toml
```diff
@@ -7,10 +7,8 @@ edition = "2021"
 moon_config = { path = "../config"}
 moon_error = { path = "../error"}
 moon_logger = { path = "../logger"}
-moon_project = { path = "../project"}
```
milesj, line 10: Pulling the cache engine into the project crate caused a cycle, so I moved the hasher to its own crate.

crates/project/src/errors.rs
```diff
@@ -15,12 +15,6 @@ pub enum ProjectError {
     )]
     InvalidConfigFile(String, String),
 
-    #[error("Failed to parse and open <file>{0}/package.json</file>: {1}")]
```
milesj, line 18: This was causing double error messages, just pass the original through.

PR #137, opened 2022-06-12
ci: Refactor integration tests to improve code coverage.

milesj:
Tests that utilize `assert_cmd` do not capture code coverage, as it's ran as a child process?

This refactors some integration tests to simply run the functions directly instead of using that crate.


milesj:
Going with this instead: https://github.com/moonrepo/moon/pull/140

PR #151, opened 2022-06-19
[feature] Add `--logFile` for appending logs to a file.

tbezman:
### Goals

- [x] Pipe logs to a file specified by the user
- [x] Ensure the log file doesn't have the color codes as it makes sifting through the logs impossible

### Notes

- coloring of logs happens all over the repo, seemed like the most sane way to remove the ansi codes was to just strip them out rather than updating the callsites.
- I didn't implement the "first log gets the full date format" stuff since it seemed unnecessary and introduced ugly complexity (if you want lemme know)
- It seems like the `--log` option wasn't working before. I fixed this by adding the `.level(level)` on the dispatcher.


crates/logger/Cargo.toml
```diff
@@ -9,3 +9,4 @@ console = "0.15.0"
 dirs = "4.0.0"
 fern = "0.6.1"
 log = "0.4.17"
+strip-ansi-escapes = "0.1.1"
```
milesj, line 12: It looks like the `console` crate we're already using has a function for this: https://docs.rs/console/0.15.0/console/fn.strip_ansi_codes.html

crates/logger/src/logger.rs
```diff
@@ -59,8 +65,40 @@ impl Logger {
...
+
+                    let message_without_colors = std::str::from_utf8(&message_without_colors_bytes)
+                        .expect("could not decode de-colored log bytes into a str");
+
+                    out.finish(format_args!("{}", message_without_colors))
+                })
+                .chain(log_file(output).expect("yikes"));
+
+            Dispatch::new()
+                .level(level)
```
milesj, line 91: My original thinking was that `--logFile` would override console logging entirely.

I'm not sure how other tools go about this, will need to look into.

milesj:
Can you also updates the options list here: https://github.com/moonrepo/moon/blob/master/website/docs/commands/overview.mdx

milesj:
@tbezman Looks good. Can you create an integration test that verifies it works. Here's an example: https://github.com/moonrepo/moon/blob/master/crates/cli/tests/run_test.rs#L130

crates/logger/src/logger.rs
```diff
@@ -59,8 +66,33 @@ impl Logger {
                     message
                 ));
             })
-            .chain(Dispatch::new().level(level).chain(io::stderr()))
-            .apply()
-            .unwrap();
+            .chain(io::stdout());
```
tbezman, line 69: @milesj this used to be stderr. that felt weird to me but idk what i'm talking about with this shit

milesj: Aww yeah, that was correct, I missed this.

Logs are written to stderr so they can be piped/captured separately from normal output.

tbezman: weird

milesj:
@tbezman Thanks for this. I'll merge it and fix the failing test in the develop branch.

PR #192, opened 2022-07-10
internal: Migrate to `AsRef`.

milesj:
All of our shared code/utils should be using `AsRef`, so this starts that initial process.


ofek:
Hey! What's the benefit of this?

milesj:
@ofek It allows you to pass `String` and `&str` to an argument.

ofek:
Oh and passing `String` would then not be a move?

milesj:
@ofek It unfortunately does not, but allows all of these patterns to "just work".

```
test("foo");
test(String::from("bar"));
test(&baz);
```

ofek:
Got it, thank you!!!

PR #239, opened 2022-08-05
new: Store task outputs as `.tar.gz` files.

milesj:
Currently, when we cache a task's outputs, we recursively copy all the files/folders into `.moon/cache/out`, and do the reverse when hydrating on a cache hit. This isn't as performant as it could be, so instead we'll now create a `tar.gz` archive for each output, and store it at `.moon/cache/out/<hash>.tar.gz`. When hydrating, we'll simply unpack the archive into the target.

This also helps simplify our infrastructure for the future remote cache mechanism.

### DIFF
--- a/crates/archive/Cargo.toml
+++ b/crates/archive/Cargo.toml
@@ -6,10 +6,8 @@ edition = "2021"
 [dependencies]
 moon_error = { path = "../error" }
 moon_logger = { path = "../logger" }
+moon_utils = { path = "../utils" }
 flate2 = "1.0.24"
 tar = "0.4.38"
 thiserror = "1.0.31"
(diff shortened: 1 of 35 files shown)


crates/archive/Cargo.toml
```diff
@@ -0,0 +1,14 @@
...
+name = "moon_archive"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+moon_error = { path = "../error" }
+moon_logger = { path = "../logger" }
+flate2 = "1.0.24"
+tar = "0.4.38"
+zip = "0.6.2"
```
milesj, line 11: add an `ArchiveError`

PR #242, opened 2022-08-06
new: Add an `outputStyle` option to  tasks.

jacobwgillespie:
### Background

I would like to be able to run multiple `dev`-style tasks at the same time - for example, I have a Remix application and run both a Remix dev server and a Tailwind CSS dev server simultaneously. With moon, that looks something like this:

```yaml
tasks:
  dev:
    command: noop
    deps:
      - app:dev-css
      - app:dev-remix

  dev-css:
    command: tailwindcss
    args: -w -i styles/app.css -o app/styles/app.css

  dev-remix:
    command: remix
    args: dev
```

Currently though, moon only streams the output of the primary task, so running `moon run app:dev` runs both dev servers, but prints nothing.

### Change

This PR adds a new task option named `streamOutput` that when true, streams the output of the task, even if it is not the primary task:

```yaml
  dev:
    command: noop
    deps:
      - app:dev-css
      - app:dev-remix

  dev-css:
    command: tailwindcss
    args: -w -i styles/app.css -o app/styles/app.css
    options:
      streamOutput: true

  dev-remix:
    command: remix
    args: dev
    options:
      streamOutput: true
```

## Review Notes

I'm not sure if this is the best option name or if this fits in with your plans for how moon works internally - let me know if anything should be changed, and no worries about rejecting this PR if it isn't the approach you would prefer. 👍

### DIFF
--- a/crates/config/src/project/task.rs
+++ b/crates/config/src/project/task.rs
@@ -94,6 +94,9 @@ pub struct TaskOptionsConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub run_from_workspace_root: Option<bool>,
+
+    #[serde(rename = "streamOutput", skip_serializing_if = "Option::is_none")]
+    pub stream_output: Option<bool>,
 }
 
 // We use serde(default) here because figment *does not* apply defaults
--- a/crates/config/src/project/task.rs
+++ b/crates/config/src/project/task.rs
@@ -65,6 +65,13 @@ pub enum TaskMergeStrategy {
     Replace,
 }
 
+#[derive(Clone, Debug, Deserialize, JsonSchema, PartialEq, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum TaskOutputStyle {
+    OnExit,
+    Stream,
+}
+
 #[derive(Clone, Debug, Default, Deserialize, JsonSchema, PartialEq, Serialize, Validate)]
 #[serde(default, rename_all = "camelCase")]
 pub struct TaskOptionsConfig {
@@ -86,6 +93,9 @@ pub struct TaskOptionsConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub merge_outputs: Option<TaskMergeStrategy>,
 
+    #[serde(rename = "outputStyle", skip_serializing_if = "Option::is_none")]
+    pub output_style: Option<TaskOutputStyle>,
+
     #[serde(skip_serializing_if = "Option::is_none")]
     pub retry_count: Option<u8>,
 
@@ -94,9 +104,6 @@ pub struct TaskOptionsConfig {
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub run_from_workspace_root: Option<bool>,
-
-    #[serde(rename = "streamOutput", skip_serializing_if = "Option::is_none")]
-    pub stream_output: Option<bool>,
 }
 
 // We use serde(default) here because figment *does not* apply defaults
--- a/crates/config/src/project/task.rs
+++ b/crates/config/src/project/task.rs
@@ -93,7 +93,7 @@ pub struct TaskOptionsConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub merge_outputs: Option<TaskMergeStrategy>,
 
-    #[serde(rename = "outputStyle", skip_serializing_if = "Option::is_none")]
+    #[serde(skip_serializing_if = "Option::is_none")]
     pub output_style: Option<TaskOutputStyle>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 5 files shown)


milesj:
@jacobwgillespie I was curious when this request was going to pop up, as I've been thinking about the best way to handle it [since v0.1](https://github.com/moonrepo/moon/blob/master/ROADMAP.md#cli). I have a few ideas, would love to hear your feedback on what you think seems best (or maybe all of them, or none of them):

1. Add a `moon run-many` command, so that that you can run concurrent tasks, like `moon run-many app:dev-css app:dev-remix`. Or even just update `moon run` to support multiple targets. The action runner has already been [designed to support it](https://github.com/moonrepo/moon/blob/master/crates/cli/src/commands/run.rs#L77).
	- Pros: Uncoupled tasks can be ran concurrently.
	- Cons: You would need to explicitly pass each target on the command line. So you couldn't just do `moon run dev`.
2. Add a `moon run --outputStyle stream` type of argument, which controls the output of all current tasks in the graph. It would support other styles like "pipe", "error-only", "hash", "none", etc.
	- Pros: More control of how running tasks are displayed.
	- Cons: May clutter the terminal. You may also not want this for _all_ tasks, just a few.
3. Add an `outputStyle` option to tasks. This would work in a similar fashion to your PR, but coupled with #2.
	- Pros: Granular control on a task level.
	- Cons: No way to disable it in other scenarios.

jacobwgillespie:
Nice 💥 

> a `moon run-many` command

I can see this being useful, though in the `dev` case I'm imagining, it's perhaps more accurate that I want to see the output of specific tasks no matter where they fall in the dependency tree. I would imagine I could `run-many x y z` and moon would "solve" any dependencies those tasks had on each other, but it doesn't exactly map to the mental model of what I'm wanting to do.

> a `moon run --outputStyle stream` type of argument

That could also work, and I do like the other styles.

Thinking about it though, moon currently prints the output of task dependencies after they have completed execution. In this case, the tasks never complete, since they are dev servers, and so no output is printed. This PR tries to solve that via an imperative option (`streamOutput`), but an alternative would be a more declarative option indicating that you expect the task to be long-running.

If there was a way to declaratively indicate that we expect the task to run continuously (`kind: long-running`, `kind: server`, idk I'm not thinking of a good name at the moment), then you could use that to determine how the output was printed. So by default maybe it's streamed, but if you did `moon run --outputStyle none` in the future it wouldn't, etc. Basically with the declarative "this task is of type X" then you could make decisions on output style afterwards.

> an `outputStyle` option to tasks

I agree with your con, this feels more rigid, and there are probably combinations of task output styles that don't work well together.

milesj:
@jacobwgillespie A "long running" option isn't a bad idea, since there's already 2 other options (3 with this) that would be affected by it. I'll look into it for the next version.

For now, let's go with option 3. If you could rename this to `outputStyle` and then create an enum for the choices (only the 2 basic ones for now). We could then move forward on this.

```rust
enum TaskOutputStyle {
  Pipe, // or Capture?
  Stream, // or Live?
}
```

Also open to what kind of naming makes sense for these from a consumer standpoint.

jacobwgillespie:
Nice, I've updated the PR to rename the option to `outputStyle` and use an enum. 👍 A couple notes:

1. I named the options `stream` and `on-exit`. Just my impressions, but `capture` makes me wonder why the output wouldn't be captured, like my mental model is that `moon` "captures" the output and replays it when there's a cache hit. And `pipe` makes me think of unix pipes, as in sending the output to somewhere else for further processing. 

   `on-exit` is very descriptive of _when_ the output should be printed, but might not fit with your future plans of other options, let me know if you'd like me to update the names.

4. The code is a little different than some of the other options and enums as the new `enum TaskOutputStyle` doesn't have a default. I _think_ this makes sense, as if the option is unspecified, then the output is streamed or not streamed based on other factors, so the absence of a concrete option seems relevant. I've flowed the value through as an `Option<TaskOutputStyle>` rather than calculating a default like the other options. Then the `stream_output` can be calculated with a match:

   ```rust
   let stream_output = match self.task.options.output_style {
       Some(TaskOutputStyle::Stream) => true,
       Some(TaskOutputStyle::OnExit) => false,
       None => is_primary || is_real_ci,
   };
   ```

   If that doesn't make sense to you, let me know and I can introduce a default value. Maybe something like `TaskOutputStyle::Unspecified`? I thought about that, but it felt like reinventing `Option<TaskOutputStyle>`.

crates/config/src/project/task.rs
```diff
@@ -86,6 +93,9 @@ pub struct TaskOptionsConfig {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub merge_outputs: Option<TaskMergeStrategy>,
 
+    #[serde(rename = "outputStyle", skip_serializing_if = "Option::is_none")]
```
milesj, line 96: Rename isn't necessary since we set it on the struct.

milesj:
@jacobwgillespie Looking good. Just 1 comment.

Using `Option` is also best, as we don't have a default state. Good choice there.

crates/config/src/project/task.rs
jacobwgillespie: Nice, fixed 👍

milesj:
@jacobwgillespie Awesome, thanks for the contribution 🙏 

This will land as part of v0.10.

PR #268, opened 2022-08-15
Add Arm Linux packages, build all Linux binaries with Docker/Depot in CI

jacobwgillespie:
This PR adds `arm64` / `aarch64` binaries for Linux, supporting both glibc and musl. This should allow using moon inside Dockerfiles built on M1 Macs, for instance.

Additionally, this PR uses an optimized Dockerfile for building all Linux builds (both x86 and arm64, for both gnu and musl). In order to avoid QEMU, it uses Depot builders to run the Arm builds on real Arm machines. ✨ 

The Dockerfile is optimized for caching all dependency crates on their own layer using [cargo-chef](https://github.com/LukeMathWalker/cargo-chef). This should speed up the Docker build when third-party dependencies are not changed.

A few notes:

1. The release profile config is moved into `.cargo/config.toml` - `cargo-chef` modifies the root `Cargo.toml` while building the dependencies, so unless the release profile config is moved elsewhere, it won't apply during dependency caching and will bust the cache
2. The Dockerfile uses [messense/rust-musl-cross](https://github.com/messense/rust-musl-cross) as the environment for statically compiling `moon`, as that image supports both amd64 and arm64. The image does _not_ supply openssl source, and instead they recommend using the `openssl-src` crate instead. We can do that here by enabling the reqwest `native-tls-vendored` feature
3. There are comments in the Dockerfile, but the basic idea is that you can select which binary you want to build via the `--target` flag, which will be built for the current `--platform` and placed on an empty `scratch` stage. Then via Docker's `--output type=local,dest=.` flag, you can copy that built binary to the current directory. The Dockerfile arranges the output layer to preserve the `target/TARGET/release/moon` structure, so it can unpack in the expected layer
4. Depot authentication happens via GitHub OIDC (the `id-token: write` permission), so no API token secrets are needed

There's not a clean way to test this in GitHub actions without just running it - I've run the builds locally via the `depot` CLI to verify that they work.


scripts/Dockerfile
```diff
@@ -0,0 +1,85 @@
+# NOTE: this Dockerfile is primarily used to compile moon's linux binaries.
+# It supports compiling x86_64 and aarch64 binaries for both glibc and musl.
+#
+# Usage: docker buildx build --output type=local,dest=./output --platform linux/amd64,linux/arm64 -f scripts/Dockerfile .
+
+# glibc
+
+FROM rust:1.62.0 AS base-gnu
```
milesj, line 8: I recently upgraded to 1.63

scripts/Dockerfile
```diff
@@ -0,0 +1,85 @@
...
+RUN cargo install cargo-chef --version ^0.1
+WORKDIR /app
+COPY .cargo .cargo
+COPY rust-toolchain.toml .
+RUN rustup update
+
+FROM base-gnu AS plan-gnu
+COPY . .
```
milesj, line 16: Should this maybe just be the crates folder and Cargo.* files?

milesj:
This is an absolutely fantastic PR. Thanks for doing this, much appreciated. Very stoked to be using Depot!

I left a few comments, but this PR is basically perfect.

scripts/Dockerfile
jacobwgillespie: Nice, updated 👍

scripts/Dockerfile
jacobwgillespie: Updated 👍 

There's also a `.dockerignore` file that's excluding quite a bit from `COPY . .`, but if you expect only the `Cargo*` and `crates` to affect the build, that's much less likely to bust the cache.

One note about `cargo chef prepare`, the trick there is that it will produce a JSON plan file, and then _only_ that plan file is copied to the next stage, so only the contents hash of the plan file affect the cache key, so it's actually okay for the COPY above to invalidate, it won't invalidate the future stage, if that makes sense.

milesj: Got it, good to know 👍

PR #277, opened 2022-08-20
internal: Refactor platforms to act like plugins/adapters.

milesj:
The organization of crates wasn't ideal, and I wasn't a fan of node/system code having to be called in random places. For example, the project graph, which tightly coupled everything and was starting to introduce cycles.

I refactored this so that platforms are now treated like plugin/adapters, and are not directly coupled via crates, and instead use a contract layer. Because of this, I was also able to move the actions into their correct platform specific crate.


crates/cli/src/helpers.rs
```diff
@@ -1,10 +1,29 @@
...
 use moon_logger::color::{no_color, supports_color};
+use moon_platform_node::NodePlatform;
+use moon_platform_system::SystemPlatform;
 use moon_terminal::create_theme;
+use moon_workspace::{Workspace, WorkspaceError};
 use std::env;
 use std::time::Duration;
 
+/// Loads the workspace and registers all available platforms!
+pub async fn load_workspace() -> Result<Workspace, WorkspaceError> {
```
milesj, line 13: CLI pieces everything together!

PR #298, opened 2022-08-31
Fix typo for arm64 in targetToPackage

jacobwgillespie:
So I just discovered that the published arm64 npm packages don't actually have a `moon` binary file included in the package, and I tracked it down to a typo of mine in `targetToPackage`. 🙈 

Since the `x86_64` targets are listed last, it thankfully didn't actually publish arm packages as x86, but this typo did prevent the arm packages from actually releasing.


milesj:
@jacobwgillespie Oh wow, I completely missed this also 😲

PR #302, opened 2022-09-03
fix prettier command in example

bram209:


milesj:
@bram209 Whoops, nice catch, thank you.

PR #311, opened 2022-09-13
QOL: Add *.profraw to .gitignore

malkir:
If they shouldn't be committed, gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,5 +37,6 @@ packages/*/moon.exe
 tests/fixtures/*/.moon/cache
 tests/fixtures/*/.yarn
 
-# Debug cruft
+# Debug and profiling cruft
 *.profraw
+*profdata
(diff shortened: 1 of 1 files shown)


malkir:
If a step can be removed from a commit workflow, it's generally a good idea to just ignore those files. You can always just use `find . -name '*.perfraw'` or run your cleanup make command if you actually see them and they are annoying.

malkir:
If a step can be removed from a commit workflow, it's generally a good idea to just ignore those files. You can always just use `find . -name '*.perfraw'` or run your cleanup make command if you actually see them and they are annoying. And the profdata ignore is for anyone doing performance profiling so they don't slip through.

malkir:
> @malkir Are you running test coverage?
 Yeah was going through docs and looking for QOL improvements

PR #312, opened 2022-09-13
QOL: making test-coverage requires git>=2.28 for git's initial-step flag

malkir:
Cannot run test-coverage without git>=2.28, Ubuntu 20.04 users for example will currently only be pulling git=2.25 from apt.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ Contributions are always welcome, no matter how large or small!
 
 - Node.js >= v14.15
 - Rust >= 1.61
+- Git >= 2.28 (for `test-coverage`)
 
 ## Setup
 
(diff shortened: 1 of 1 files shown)


milesj:
@malkir What actually happens on older git versions, does it throw an error?

malkir:
Flat out fails due to that flag not existing in versions of git before 2.28

malkir:
> @malkir What actually happens on older git versions, does it throw an error?

@milesj Flat out fails due to that flag not existing in versions of git before 2.28

milesj:
@malkir Unfortunate, hah.

Also, is your branch fully rebased off master? I'm not sure why the status checks are not running.

malkir:
> @malkir Unfortunate, hah.
> 
> Also, is your branch fully rebased off master? I'm not sure why the status checks are not running.

Should be, I rebased off upstream- will verify when I get home later

PR #326, opened 2022-09-21
new: Add pruning for `Dockerfile`s.

milesj:
Continuing with supporting Docker, this adds a new command `moon docker prune`, that cleans up the container by doing the following:

- Only installs production dependencies for configured tools.
- Deletes extraneous files and folders.


crates/toolchain/src/pms/pnpm.rs
```diff
@@ -218,4 +218,30 @@ impl PackageManager<NodeTool> for PnpmTool {
...
         Ok(())
     }
+
+    async fn install_focused_dependencies(
+        &self,
+        toolchain: &Toolchain,
+        package_names: &[String],
+        production_only: bool,
+    ) -> Result<(), ToolchainError> {
+        let mut cmd = self.create_command();
```
jacobwgillespie, line 228: Does this need something like `cmd.arg("install");`? I might be missing where it adds the install command, but the other two package managers add it directly in their implementations.

milesj: Yeah nice catch.

PR #343, opened 2022-10-01
new: Support per-project overrides for tool versions.

milesj:
This is a long standing request and we finally have the infrastructure in place to support it. This change will allow each project to override the Node.js version (and versions of future tools) at the workspace-level.

This will help to support legacy projects that are stuck on an explicit version.


crates/utils/src/lib.rs
```diff
@@ -24,6 +25,17 @@ macro_rules! string_vec {
     }};
 }
 
+#[cached]
+pub fn get_workspace_root() -> PathBuf {
```
milesj, line 29: This is super handy so we don't have to pass it around as an argument or struct field, which got very convoluted.

PR #380, opened 2022-10-19
Fix size of terminal on mobile screens

IgnisDa:
The existing `HeroTerminal` did not have a fixed height on smaller screens. This resulted in a layout shift once the text appeared.

image

This PR gives it a fixed height on all screens.

image


milesj:
@IgnisDa Thanks!

PR #384, opened 2022-10-21
Check if repository is dirty before migrating

IgnisDa:
This PR adds a simple check to see whether there are any dirty/modified files in the project before continuing to migrate. This will prevent unintentional loss of changed files.

Also a small refactor and moved the `detect_vcs` version to the `moon_vcs` library.

### DIFF
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -400,7 +400,8 @@ pub enum Commands {
     // moon migrate <operation>
     #[command(
         name = "migrate",
-        about = "Operations for migrating existing projects to moon."
+        about = "Operations for migrating existing projects to moon.",
+        rename_all = "camelCase"
     )]
     Migrate {
         #[command(subcommand)]
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -20,13 +20,13 @@ const HEADING_PARALLELISM: &str = "Parallelism and distribution";
 pub enum DockerCommands {
     #[command(
         name = "prune",
-        about = "Remove extraneous files and folders within Dockerfiles."
+        about = "Remove extraneous files and folders within Dockerfile(s)."
     )]
     Prune,
 
     #[command(
         name = "scaffold",
-        about = "Scaffold a repository skeleton for use within Dockerfiles."
+        about = "Scaffold a repository skeleton for use within Dockerfile(s)."
     )]
     Scaffold {
         #[arg(required = true, help = "List of project IDs to copy sources for")]
@@ -390,7 +390,7 @@ pub enum Commands {
     // moon docker <operation>
     #[command(
         name = "docker",
-        about = "Operations for integrating with Docker and Dockerfiles."
+        about = "Operations for integrating with Docker and Dockerfile(s)."
     )]
     Docker {
         #[command(subcommand)]
@@ -405,6 +405,9 @@ pub enum Commands {
     Migrate {
         #[command(subcommand)]
         command: MigrateCommands,
+
+        #[arg(long, help = "Disable the check for touched/dirty files")]
+        skip_touched_files_check: bool,
     },
 
     // moon query <operation>
--- a/crates/cli/src/commands/clean.rs
+++ b/crates/cli/src/commands/clean.rs
@@ -1,7 +1,7 @@
 use crate::helpers::{create_progress_bar, load_workspace};
 
 pub struct CleanOptions {
-    pub cache_liftime: String,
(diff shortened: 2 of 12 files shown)


IgnisDa:
Here is the check in action:
image

crates/cli/src/app.rs
```diff
@@ -405,6 +405,9 @@ pub enum Commands {
     Migrate {
```
milesj, line 405: Add `rename_all = "camelCase"` to the `#[command]` above

crates/vcs/src/lib.rs
```diff
@@ -4,8 +4,48 @@ mod loader;
...
+use strum::Display;
+
 pub use errors::VcsError;
 pub use git::Git;
 pub use loader::*;
 pub use svn::Svn;
 pub use vcs::*;
+
+#[derive(Debug, Display)]
+pub enum SupportedVcs {
```
milesj, line 17: Let's use this enum instead of creating another https://github.com/moonrepo/moon/blob/master/crates/config/src/workspace/vcs.rs#L7

crates/vcs/src/lib.rs
```diff
@@ -4,8 +4,48 @@ mod loader;
...
+        return Ok((
+            SupportedVcs::Svn,
+            Svn::new("trunk", dest_dir).get_local_branch().await?,
+        ));
+    }
+    Ok((SupportedVcs::Git, "master".into()))
+}
+
+/// Get all the touched/dirty files in the repository
+pub async fn get_touched_files(path: &Path) -> Result<HashSet<String>, Box<dyn std::error::Error>> {
```
milesj, line 44: If possible, I'd prefer to use the vcs found on the workspace instance here: https://github.com/moonrepo/moon/blob/master/crates/workspace/src/workspace.rs#L119 Since this is the vcs that's configured in `.moon/workspace.yml`.

However, you may have to defer the dirty checks until inside the `from_package_json` call.

crates/cli/src/commands/migrate/mod.rs
```diff
@@ -1,3 +1,11 @@
 mod from_package_json;
 
 pub use from_package_json::from_package_json;
+use moon_workspace::Workspace;
+
+pub async fn check_dirty_repo(workspace: &Workspace) -> Result<(), Box<dyn std::error::Error>> {
+    if !workspace.vcs.get_touched_files().await?.all.is_empty() {
+        Err("Commit or stash your changes before running this command, or use the `--skipTouchedFilesCheck` flag to disable this check.".to_string())?;
```
milesj, line 8: missing `return`

crates/cli/src/lib.rs
```diff
@@ -132,9 +133,32 @@ pub async fn run_cli() {
...
             .await
         }
-        Commands::Migrate { command } => match command {
-            MigrateCommands::FromPackageJson { id } => migrate::from_package_json(id).await,
-        },
+        Commands::Migrate {
+            command,
+            skip_touched_files_check,
+        } => {
+            let maybe_workspace = load_workspace().await;
```
milesj, line 140: does `?` not bubble up here?

IgnisDa: Nope it does not:
image

crates/cli/src/commands/migrate/mod.rs
IgnisDa: AFAIK the `?` does that. Added the return though to make it more explicit.

crates/cli/src/lib.rs
milesj: Hrmmm. There's a lot going on here. Maybe just pass the flag t the method like:

```
migrate::from_package_json(workspace, id, *skip_touched_files_check).await)
```

IgnisDa: You are right, I moved both the `load_workspace` and `check_dirty_repo` inside the `from_package_json` function. Looks much cleaner now.

crates/cli/src/commands/clean.rs
```diff
@@ -1,20 +1,20 @@
 use crate::helpers::{create_progress_bar, load_workspace};
 
 pub struct CleanOptions {
-    pub cache_liftime: String,
+    pub cache_lifetime: String,
```
milesj, line 4: Nice catch!

milesj:
@IgnisDa Looks good, you'll just need to fix the version check by running `yarn version:bump:bin patch` and committing the file.

IgnisDa:
@milesj Done!

milesj:
Awesome!

Issue #144, opened 2022-06-19
[feature] Apply `#[inline]` to Rust functions


milesj:
**Is your feature request related to a problem? Please describe.**

Nope

**Describe the solution you'd like**

We should apple `#[inline]` to our functions (when applicable) for performance improvements. https://nnethercote.github.io/perf-book/inlining.html

**Describe alternatives you've considered**

**Additional context**

I've never used inlining in Rust before, so I'm not sure what "best practices" are. I did notice that Rome is using it heavily: https://github.com/rome/tools

PR #387, opened 2022-10-23
perf: Enable Rust inlining.

milesj:
Implements https://github.com/moonrepo/moon/issues/144

I primarily enabled `#[inline]` for the utils crate, since its used everywhere.


milesj:
!benchmark

PR #388, opened 2022-10-23
internal: Rework Rust lifetimes.

milesj:
Trying to go with a composition approach as it reduces the amount of stuff to copy/pass around.


milesj:
This got way too complicated.

PR #399, opened 2022-10-29
Add a `--all` flag for `check` subcommand

### LINKED ISSUE #398: Run `check` on all projects if no project is specified
I am writing a pre-push hook that builds and tests all projects. The `moon check` is the perfect command for this. However I have to list all the projects projects manually. This is error prone and I have to remember to update the command everytime I add a project.

I propose a `--all` flag to the `moon check` command that will run the command on all the projects configured in the `.moon/workspace.yml` file.

IgnisDa:
Fixes #398.

### DIFF
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -289,10 +289,15 @@ pub enum Commands {
     )]
     Check {
         #[arg(help = "List of project IDs to explicitly check")]
+        #[clap(group = "projects")]
         ids: Vec<ProjectID>,
 
         #[arg(long, help = "Generate a run report for the current actions")]
         report: bool,
+
+        #[arg(long, help = "Run check for all projects in the workspace")]
+        #[clap(group = "projects")]
+        all: bool,
     },
 
     // moon ci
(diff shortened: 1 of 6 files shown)


milesj:
@IgnisDa I think an `--all` option would be better, otherwise this wouldn't work when not ran from the root.

IgnisDa:
@milesj I did not think of that possibility. I added a `--all` flag and added 2 tests for it. Please review.

crates/cli/src/commands/check.rs
```diff
@@ -1,23 +1,36 @@
...
 
 pub struct CheckOptions {
     pub report: bool,
 }
 
+const LOG_TARGET: &str = "moon:check";
+
 pub async fn check(
     project_ids: &Vec<String>,
+    all: bool,
```
milesj, line 15: can you put the all in `CheckOptions`

crates/cli/src/app.rs
```diff
@@ -289,10 +289,15 @@ pub enum Commands {
     )]
     Check {
         #[arg(help = "List of project IDs to explicitly check")]
+        #[clap(group = "projects")]
```
milesj, line 292: what does this do exactly?

IgnisDa: This makes `[ID]s` and `--all` mutually exclusive so that they can not be used together.

IgnisDa: image

milesj: Awesome

PR #410, opened 2022-11-01
fix: Fix stdout/stderr logs not existing.

milesj:
It's weird that a command would not produce these files, but let's check for this anyways.


milesj:
!benchmark

PR #412, opened 2022-11-02
internal: Migrate from std HashMap/HashSet to FxHashMap/FxHashSet.

milesj:
More information: https://nnethercote.github.io/perf-book/hashing.html


milesj:
!benchmark

PR #422, opened 2022-11-07
Interactive dep graph

### LINKED ISSUE #10: [feature] Interactive dep/project graph

No, new feature.

Currently the `dep-graph` and `project-graph` commands output DOT that can be piped to a file. Instead we should open a browser window that renders an interactive project graph will full data.

The DOT output can be enabled with a `--dot` arg.

N/A

**Additional context**

IgnisDa:
Fixes #10.

Information about how it works is written in the visualizer `README`.


IgnisDa:
The frontend is still very basic but it is capable of displaying the project graph:

image

milesj:
@IgnisDa I appreciate the effort, but using Rocket and GraphQL is entirely overkill. I'm almost positive this wouldn't work in editor extensions either.

For serving the HTML/JavaScript, all we need is something like this: https://github.com/tiny-http/tiny-http

And for data, we don't need GQL or any kind of API. We have all the data upfront, theres nothing to query, we can just inject it into the HTML as a JSON blob.

apps/visualizer/README.md
```diff
@@ -0,0 +1,26 @@
+# Visualizer
```
milesj, line 1: This would be a package, not an app, since it needs to be published to npm.

apps/visualizer/plugins/createGitignore.ts
```diff
@@ -0,0 +1,24 @@
+import { writeFileSync } from 'fs';
+import type { PluginOption } from 'vite';
+import { normalizePath } from 'vite';
+
+/**
+ * The `dist` directory is deleted every time the bundle is rebuilt. The `.gitignore` file
```
milesj, line 6: The dist folder doesn't need to be in git, just in the npm package when published.

crates/cli/src/app.rs
```diff
@@ -186,6 +186,12 @@ pub enum Commands {
     )]
     Teardown,
 
+    #[command(
+        name = "visualize",
+        alias = "vis",
+        about = "Start an interactive monorepo visualization graph"
+    )]
+    Visualize,
```
milesj, line 194: The visualizer would be for `moon dep-graph` and `moon project-graph`, we shouldn't add a 3rd command.

IgnisDa: Alright makes sense. I thought those were reserved for the dot versions.

apps/visualizer/plugins/createGitignore.ts
IgnisDa: Alright

IgnisDa:
@milesj That makes sense. So if I am getting this right, you want the data to be injected into the HTML (possibly by `tera`) and then the JS would render the graph in the browser. 

Where do you think I should write the JS for this? In a separate package or just in the rust code itself?

milesj:
@IgnisDa We can use https://unpkg.com/ to include the JS on the Rust side.

I think it may be best to split the Rust and JS code in separate PRs. So maybe for this PR, it's just building the visualizer JS side of things?

IgnisDa:
Alright will make a separate PR. Closing this.

Issue #414, opened 2022-11-02
[bug] fitler on change status does not work


tomdavidson:
**Describe the bug**

Exploring moon's filtering to replace lint-staged githook and want to filter based on the status of the files changed. Unfortunately, the filtering on change status does not work.

**Steps to reproduce**

1. Create a dummy root level task in //moon.yml
```yaml
  ls:
    command: ls
    platform: system
```

2. Make sure there are changed files in git with various status such as modified or staged.

3. Run the task with the status filter as documented `$ moon run root:ls --status staged`

**Expected behavior**

The status filtering on staged should match the files that git shows are staged. 

**Environment** 

  System:
    OS: Linux 5.11 Ubuntu 21.04 (Hirsute Hippo)
    CPU: (4) x64 Intel(R) Core(TM) i5-7300U CPU @ 2.60GHz
    Memory: 17.67 GB / 30.75 GB
    Container: Yes
    Shell: 5.1.4 - /bin/bash
  Binaries:
    Node: 18.11.0 - /usr/bin/node
    npm: 8.15.0 - ~/.local/bin/npm
  Managers:
    Apt: 2.2.4 - /usr/bin/apt
    Cargo: 1.64.0 - ~/.cargo/bin/cargo
    pip3: 20.3.4 - /usr/bin/pip3
  Utilities:
    Make: 4.3 - /usr/bin/make
    GCC: 10.3.0 - /usr/bin/gcc
    Git: 2.30.2 - /usr/bin/git
    FFmpeg: 4.3.2 - /usr/bin/ffmpeg
  Virtualization:
    Docker: 20.10.14 - /usr/bin/docker
    VirtualBox: 6.1.26 - /usr/bin/vboxmanage
  IDEs:
    Nano: 5.4 - /usr/bin/nano
    VSCode: 1.72.2 - /usr/bin/code
  Languages:
    Bash: 5.1.4 - /usr/bin/bash
    Go: 1.16.6 - /usr/bin/go
    Perl: 5.32.1 - /usr/bin/perl
    Python3: 3.9.5 - /usr/bin/python3
    Rust: 1.64.0 - /home/tom/.cargo/bin/rustc
  Browsers:
    Firefox: 96.0
</details>

**Additional context**
*<summary>terminal output</summary>*
<details>
```shell
$ git status
On branch ci/tune-up-eslint
Your branch is up to date with 'origin/ci/tune-up-eslint'.

Changes to be committed:
[...]

PR #423, opened 2022-11-07
internal: Clean up and test affected logic.

milesj:
Investigation into https://github.com/moonrepo/moon/issues/414


milesj:
!benchmark

PR #452, opened 2022-11-17
new: Support file system diffing when hydrating outputs.

milesj:
When hydrating outputs from a tarball, we now compare the dest and source files to determine whether to write the source or not. Previously we would just wipe all files before hand, which wasn't very performant. This change has resulted in 10x the speed.

time:   [226.88 ms 235.45 ms 244.67 ms]
time:   [27.861 ms 28.181 ms 28.542 ms]


milesj:
!benchmark

PR #457, opened 2022-11-18
new: Add `MOON_AFFECTED_FILES` env var and improve flow.

milesj:
Fixes some edge cases and improves the affected flow.

### DIFF
--- a/crates/cli/tests/run_node_test.rs
+++ b/crates/cli/tests/run_node_test.rs
@@ -1321,15 +1321,29 @@ mod affected_files {
             .assert();
         let output = get_assert_output(&assert);
 
-        // Order is not deterministic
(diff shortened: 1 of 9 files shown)


tomdavidson:
If the task command uses the env var, the affected files are still appended to the args list?

What about something such as `--affected` sets MOON_AFFECTED_FILES to the files rather than a default of `.` and ` options.affectedFiles: true` appends the args list?

Separately, will MOON_AFFECTED_FILES persist to task dependencies?

milesj:
@tomdavidson I have an idea but i'll have to wait till 0.20. Basically, I'll update the setting to support an enum, like `affectedFiles: 'args'` or `affectedFiles: 'env'`.

PR #461, opened 2022-11-22
internal: Re-use `Target` instance everywhere.

milesj:
I found myself constantly parsing the target IDs, so this keeps and uses the `Target` instance around.


milesj:
!benchmark

PR #484, opened 2022-12-01
new: Refactor project and dependency graphs.

milesj:
When I first built the project graph, I built it to:

- Lazy load projects on demand, instead of eager loading everything upfront.
- Have the graph available from the workspace instance, since it'll be needed everywhere.

However, this has caused _a ton_ of problems, primarily with Rust's lifetime and borrow system. The biggest issue is that lazy loading graph nodes requires the graph to be mutable, and in turn, the workspace. I worked around this with interior mutability, but this resulted in _a fuck ton_ of cloning, as I wasn't able to return project's as refs. This is definitely a performance hit.

So in the end, I said screw it, and I rewrote the project graph to eager load all projects up front. This avoids interior mutability, and allows projects as refs... no more cloning! It's also no longer coupled to the workspace, which allows the workspace to be in read-only more for the most part. Big win here.

While I was refactoring the project graph, I was also able to achieve 2 more improvements:

- Moved the platforms layer to the workspace. This really needed to be at the top-level, not in the project graph.
- Refactored the dependency graph to work like the new project graph. We're also able to pass references through the constructor now using lifetimes, which is a huge win.


milesj:
!benchmark

PR #486, opened 2022-12-02
internal: Make all fs calls synchronous.

milesj:
When I started building moon, I made all fs calls async as I was used to Node.js. However, in Rust, they work quite differently. Since OS fs calls are sync by default, Tokio fs is merely an async wrapper that _polls_ until complete. I'm not 100% positive, but I have a hunch this adds a non-trivial amount of processing cost.

This changes all fs calls to sync, which in turn converted a ton of async functions to sync. In theory, this should increase performance as we're no longer overloading the tokio runtime with a ton of useless async calls.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -106,17 +106,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-recursion"
-version = "1.0.0"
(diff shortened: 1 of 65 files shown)


milesj:
!benchmark

PR #497, opened 2022-12-11
internal: Rewrite task expansion and token resolving.

milesj:
There's a massive bug with `^:` dependencies, as it would resolve a task for all parents, not just for those that have a matching task. This was impossible to fix in previous versions, but with the project graph refactor it's now possible.

To resolve this correctly, we had to move task expansion int the project graph, which in turn, also pulled token resolving. In the end, this actually made more logical sense as the project graph is "building" all of these parts.


crates/core/platform/src/lib.rs
```diff
@@ -34,14 +34,20 @@ pub trait Platform: Debug + Send + Sync {
...
         aliases_map: &mut ProjectsAliasesMap,
     ) -> Result<(), MoonError> {
         Ok(())
     }
 
-    /// During project creation (when being lazy loaded and instantiated in the graph),
-    /// scan for any implicit project dependency relations using the platforms manifest.
+    /// During project creation within the project graph, find a matching alias
+    /// for the previously loaded map of aliases, if applicable
+    fn load_project_alias(&self, aliases_map: &ProjectsAliasesMap) -> Option<String> {
```
milesj, line 45: TODO

PR #505, opened 2022-12-16
new: Add basic support for new languages: Go, PHP, Python, Ruby, Rust

milesj:
Now that our tooling supports this, we can start adding new languages.


milesj:
!benchmark

PR #512, opened 2022-12-21
internal: Clean up platform + tool integration.

milesj:
I'm not happy with how coupled some of these crates are, mainly the toolchain and platform layers. The toolchain existed far before the platform concept, so a lot of that is coupled to the workspace. In the future, this will change as the tool will be within the platform, but until we get to that point, I need to start decoupling things.

The biggest change in this PR is that tools are now their own crates, and are pieced together using `Any` magic.


milesj:
!benchmark

PR #518, opened 2022-12-27
new: Implement new pipeline for processing actions.

milesj:
This adds a new crate, the action pipeline, that is now in charge of executions actions (tasks in the dep graph). The pipeline uses a new worker pool approach so that we can control the amount of concurrency. This isn't performance tuned yet, as I will continually work on this in the branch.

Furthermore, it generalizes many actions as they'll be the same across all languages. With this approach, language specific functionality is now implemented in the platform!


milesj:
!benchmark

PR #547, opened 2023-01-08
internal: Migrate to new artifacts endpoints.

milesj:
We'll be utilizing presigned URLs entirely for this.

### DIFF
--- a/crates/core/moonbase/src/lib.rs
+++ b/crates/core/moonbase/src/lib.rs
@@ -5,7 +5,7 @@ mod errors;
 use common::{endpoint, get_request, post_request, Response};
 use moon_error::map_io_to_fs_error;
 use moon_logger::{color, debug, warn};
-use reqwest::Body;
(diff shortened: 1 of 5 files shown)


crates/core/moonbase/src/lib.rs
```diff
@@ -128,77 +147,85 @@ impl Moonbase {
...
-        reqwest::Client::new().post(url).multipart(form)
+        reqwest::Client::new()
+            .post(url)
+            .body(Body::wrap_stream(file_stream))
     } else {
         reqwest::Client::new()
-            .post(format!("{}/artifacts/{}/upload", get_host(), hash))
-            .multipart(form)
-            .bearer_auth(auth_token)
+            .post(endpoint(format!("artifacts/{}/upload", hash)))
```
jpoz, line 179: thought we were removing upload?

milesj: im going to leave the logic in for now in case we need it in the future

PR #567, opened 2023-01-26
fix: Fix project graph not taking into account untracked changes.

milesj:
When creating a new config file, the pg cache sometimes wouldn't take it into account. I noticed this when debugging `git ls-tree`, as it doesn't hash untracked files.

To work around this, I dropped the usage of `git ls-tree` and replaced it with `git ls-files` that then gets piped into `git hash-object`.


crates/core/project-graph/src/project_builder.rs
```diff
@@ -545,31 +560,32 @@ impl<'ws> ProjectGraphBuilder<'ws> {
...
-            let config_hashes = self
-                .workspace
-                .vcs
-                .get_file_hashes(
-                    &convert_paths_to_strings(&configs, &self.workspace.root)?,
-                    false,
-                )
-                .await
-                .map_err(|e| MoonError::Generic(e.to_string()))?;
+        let config_hashes = self
```
milesj, line 581: switch to hash the entire `.moon` folder.

jpoz (approved):

PR #568, opened 2023-01-26
Check for a new version on launch in the background

jpoz:
Check if there's a new version of moon

### DIFF
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -22,6 +22,7 @@ moon_dep_graph = { path = "../core/dep-graph" }
 moon_error = { path = "../core/error" }
 moon_generator = { path = "../core/generator" }
 moon_lang = { path = "../core/lang" }
+moon_launchpad = { path = "../core/launchpad" }
 moon_logger = { path = "../core/logger" }
 moon_node_lang = { path = "../node/lang" }
 moon_node_platform = { path = "../node/platform" }
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -25,10 +25,12 @@ use app::{App, Commands, DockerCommands, MigrateCommands, NodeCommands, QueryCom
 use clap::Parser;
 use console::Term;
 use enums::{CacheMode, LogLevel};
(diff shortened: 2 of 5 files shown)


crates/cli/Cargo.toml
```diff
@@ -52,11 +52,13 @@ mimalloc = { version = "0.1.34", default-features = false }
 open = "3.2.0"
 petgraph = { workspace = true }
 rustc-hash = { workspace = true }
+reqwest = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_yaml = { workspace = true }
 strum = { version = "0.24.1", features = ["derive"] }
 tera = { version = "1.17.1", features = ["preserve_order"] }
+lenient_semver = { version = "0.4.2", features = ["version_lite", "version_serde"] }
```
milesj, line 61: use `moon_utils::semver` instead

crates/cli/src/lib.rs
```diff
@@ -25,13 +25,24 @@ use app::{App, Commands, DockerCommands, MigrateCommands, NodeCommands, QueryCom
...
 use moon_terminal::ExtendedTerm;
+use serde::{Deserialize, Serialize};
 use std::env;
+use std::error::Error;
 use std::path::PathBuf;
+use tokio::task;
 
 pub use app::BIN_NAME;
 
+const CURRENT_VERSION_URL: &str = "https://launch.moonrepo.app/versions/cli/current";
```
milesj, line 39: maybe move the launchpad code to the `moonbase` crate, like `moonbase/src/launchpad.rs`?

crates/cli/src/lib.rs
```diff
@@ -73,13 +84,37 @@ fn setup_caching(mode: &CacheMode) {
     }
 }
 
+async fn check_version() -> Result<Option<String>, Box<dyn Error + Send + Sync>> {
+    let local_version_str = env!("CARGO_PKG_VERSION");
+    debug!("Checking for new version of moon");
+    let response = reqwest::Client::new()
```
milesj, line 90: should we cache this request and only fire it like every hour/day? doing it every execution seems like too much

PR #587, opened 2023-02-01
fix: Fix issues around the Node.js tool not being configured.

milesj:
Technically the tool is optional, and is optional when the `version` field in `toolchain.yml` is not set, or is set to null. When this happens, we should default to the global "node" found in the user's shell.

### DIFF
--- a/crates/node/platform/src/actions/run_target.rs
+++ b/crates/node/platform/src/actions/run_target.rs
@@ -1,6 +1,8 @@
 use crate::target_hasher::NodeTargetHasher;
 use moon_action_context::{ActionContext, ProfileType};
-use moon_config::{HasherConfig, HasherOptimization, NodePackageManager, TypeScriptConfig};
+use moon_config::{
+    HasherConfig, HasherOptimization, NodeConfig, NodePackageManager, TypeScriptConfig,
+};
 use moon_error::MoonError;
 use moon_logger::{color, trace};
 use moon_node_lang::{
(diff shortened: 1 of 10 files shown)


crates/node/platform/src/actions/run_target.rs
```diff
@@ -77,6 +79,60 @@ fn create_node_options(
...
+    let possible_bin_path = match node::find_package_bin(starting_dir, bin_name)? {
+        Some(bin) => Ok(bin),
+        None => Err(ToolError::MissingBinary(bin_name.to_owned())),
+    };
+
+    match possible_bin_path? {
+        // Rust, Go
+        BinFile::Binary(bin_path) => {
+            return Ok(Some(Command::new(bin_path)));
+        }
```
jpoz, line 98: I saw this in the original code. Why is this needed in the node package?

milesj: @jpoz Node (npm) packages must be executed with the node binary. Some npm packages ship go/rust code (like us), and trying to pass it to the node binary crashes, so we replace the node binary entirely with the go/rust one.

crates/node/platform/src/platform.rs
```diff
@@ -429,8 +430,26 @@ impl Platform for NodePlatform {
         runtime: &Runtime,
         working_dir: &Path,
     ) -> Result<Command, ToolError> {
-        let tool = self.toolchain.get_for_version(runtime.version())?;
-        let command = actions::create_target_command(tool, context, project, task, working_dir)?;
+        let version = runtime.version();
+
+        let command = if version.is_latest() {
```
jpoz, line 435: So if it's latest, it wasn't configured?

milesj: The version fallsback to the string "latest", so it's the only reference we have right now.

PR #594, opened 2023-02-03
internal: Require a subscription for remote caching.

milesj:
We'll be launching our remote caching service real real soon. This makes the final changes in preparation for that.

### DIFF
--- a/crates/core/workspace/src/workspace.rs
+++ b/crates/core/workspace/src/workspace.rs
@@ -259,7 +259,8 @@ impl Workspace {
             return Ok(());
         };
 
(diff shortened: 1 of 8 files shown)


crates/core/workspace/src/workspace.rs
```diff
@@ -259,17 +259,20 @@ impl Workspace {
             return Ok(());
         };
 
-        let Ok(api_key) = env::var("MOONBASE_API_KEY") else {
+        let Ok(access_key) = env::var("MOONBASE_ACCESS_KEY")
```
jpoz, line 262: +1 to Access Key

jpoz (approved):

Issue #526, opened 2022-12-30
[feature] `upgrade` command to the binary


itsezc:
**Is your feature request related to a problem? Please describe.**

Add the ability to upgrade the moon binary with an `upgrade` command

**Describe the solution you'd like**

The command should check for the latest release of moon and install it, additionally passing a `--nightly` flag should install a nightly version if available.

**Describe alternatives you've considered**

Re-running the install script but this is not good UX

**Additional context**

N/A

milesj:
This is done.

PR #597, opened 2023-02-04
Introduce upgrade command in CLI

nrempel:
Hey!

I'm tinkering with #526. Are you open to something like this?

This doesn't consider installs through yarn and other methods. Maybe we could hide this command behind an `upgrade` feature and not compile it for distribution to npm?

### DIFF
--- a/crates/cli/src/commands/upgrade.rs
+++ b/crates/cli/src/commands/upgrade.rs
@@ -1,0 +1,107 @@
+use crate::helpers::{create_progress_bar, AnyError};
+use bytes::Buf;
+use itertools::Itertools;
+use moon_launchpad::check_version;
+use moon_logger::error;
+use moon_utils::{
+    fs::{create_dir_all, rename},
+    semver::Version,
+};
+use proto::ProtoError;
+use std::{env, fs::File, io::copy, path::Component};
+
+pub async fn upgrade() -> Result<(), AnyError> {
+    let version = env!("CARGO_PKG_VERSION");
+    let version_check = check_version(version).await;
+
+    let new_version = match version_check {
+        Ok((newer_version, _)) if Version::parse(&newer_version)? > Version::parse(version)? => {
+            newer_version
+        }
+        Ok(_) => {
+            println!("You're already on the latest version of moon!");
+            return Ok(());
+        }
+        Err(err) => {
+            error!("Failed to get current version of the cli from remote: {err}");
+            return Err(err);
+        }
+    };
+
+    let target = match (std::env::consts::OS, std::env::consts::ARCH) {
+        ("linux", arch) => {
+            // Run ldd to check if we're running on musl
+            let output = std::process::Command::new("ldd")
+                .arg("--version")
+                .output()?;
+            let output = String::from_utf8(output.stdout)?;
+            let libc = match output.contains("musl") {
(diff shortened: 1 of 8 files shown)


milesj:
@nrempel Awesome stuff!

I'm not sure how we would solve the the npm/yarn/pnpm install situation. We're also looking into homebrew which is another can of worms.

Maybe for now, perhaps this command only works if the current exe is within the `~/.moon` folder? If not, we error and mention that you should upgrade with the tool you installed with.

crates/cli/src/commands/upgrade.rs
```diff
@@ -0,0 +1,82 @@
...
+use moon_logger::error;
+use tokio::{
+    fs::{create_dir_all, rename, File},
+    io::copy,
+};
+use tokio_util::compat::FuturesAsyncReadCompatExt;
+
+pub async fn upgrade() -> Result<(), AnyError> {
+    let version = env!("CARGO_PKG_VERSION");
+    let version_check = check_version(version).await;
```
milesj, line 15: nice reuse!

crates/cli/src/commands/upgrade.rs
```diff
@@ -0,0 +1,82 @@
...
+            let output = String::from_utf8(output.stdout)?;
+            let libc = match output.contains("musl") {
+                true => "musl",
+                false => "gnu",
+            };
+            format!("moon-{arch}-unknown-linux-{libc}")
+        }
+        ("macos", arch) => format!("moon-{arch}-apple-darwin"),
+        ("windows", "x86_64") => "moon-x86_64-pc-windows-msvc.exe".to_owned(),
+        _ => unreachable!(),
```
milesj, line 44: can return `Err(ProtoError::UnsupportedArchitecture(...))`

crates/cli/src/commands/upgrade.rs
```diff
@@ -0,0 +1,82 @@
...
+
+    let bin_path = env::current_exe()?;
+
+    // Move the old binary to a versioned path
+    let ver_path = bin_path
+        .parent()
+        .unwrap()
+        .join(&version)
+        .join(bin_path.file_name().unwrap());
+    create_dir_all(ver_path.parent().unwrap()).await?;
```
milesj, line 57: use the `moon_utils::fs` methods for most (if not all) fs operations, as it has built in io handling

nrempel:
Thanks @milesj, I patched those up. This works in my testing. Though, I only have an m1 Mac to test on.

Not sure if you want to block on some tests being added. I'm happy to add those but it might be 0-2 weeks!

milesj:
@nrempel Code looks good!

Just need to fix the unix test. As or the failing complex lockfile test, I believe I fixed that but in the `develop-0.24` branch, which will be landing on Monday. So we have 2 options here, wait for that to land and then rebase, or your rebase off the develop branch and change the base in the PR to develop as well. I'll let you decide!

nrempel:
Great! I'll rebase in a few days.

milesj:
@nrempel That PR landed

nrempel:
Rebased! Tests pass locally 🤞🏻

nrempel:
done!

nrempel:
Thank you!

PR #599, opened 2023-02-05
internal: Add docs for implementing a new language.

milesj:


scripts/new-language.md
```diff
@@ -0,0 +1,177 @@
...
+
+	#[strum(serialize = "kotlin")]
+	Kotlin,
+}
+```
+
+- [ ] Updated enum
+- [ ] Updated TypeScript types at `packages/types/src/project-config.ts`
+- [ ] Verified all `match` callsites handle the new variant
+- [ ] Ran `cargo make json-schemas` and updated the JSON schemas
```
jpoz, line 29: Don't think we have cargo make anymore

milesj: were still using cargo make in the moon repo

PR #600, opened 2023-02-06
internal: Support tracking CI runs in moonbase.

milesj:


crates/core/action-pipeline/src/processor.rs
```diff
@@ -177,17 +177,9 @@ pub async fn process_action(
         }
     };
 
-    local_emitter
```
milesj, line 180: This wasn't capturing the finished dates/final state, since it happened in the next block <_<

PR #609, opened 2023-02-10
new: Support interactive tasks.

milesj:
Interactive tasks weren't possible with our current architecture, so this implements 2 ways to solve it.


crates/cli/src/lib.rs
```diff
@@ -88,7 +88,7 @@ pub async fn run_cli() {
     let version_check = task::spawn(check_version(version));
 
     // Match and run subcommand
-    let result = match &args.command {
+    let result = match args.command {
```
milesj, line 91: changed this to avoid all the clones/derefs

PR #611, opened 2023-02-11
internal: Rework estimate calculation.

milesj:
We have a mechanism where we attempt to calculate how much time a CI pipeline would have taken when _not using moon_. It was rudimentary as it added all action durations together to generate a total duration, but this isn't how most CI pipelines work. No one is running everything serially, they run everything in parallel.

To account for this, we now do a few things:

- We ignore all actions except for run target, as we only care about the tasks themselves.
- We bucket tasks based on name, and add them together. We make an assumption that all tasks are run serially, like lint/build/format/etc.
- We take the longest bucket duration and assume this is how long their CI would take.


crates/core/action-pipeline/src/worker_pool.rs
```diff
@@ -1,67 +0,0 @@
-use async_channel::Sender;
```
milesj, line 1: wasnt used anymore

jpoz (approved):

PR #616, opened 2023-02-13
fix: Improve estimate calculation.

milesj:
Noticed some buginess while testing this.

### DIFF
--- a/crates/core/action-pipeline/src/estimator.rs
+++ b/crates/core/action-pipeline/src/estimator.rs
@@ -1,4 +1,4 @@
-use moon_action::{Action, ActionNode};
+use moon_action::{Action, ActionNode, ActionStatus};
 use moon_target::Target;
 use rustc_hash::FxHashMap;
 use serde::Serialize;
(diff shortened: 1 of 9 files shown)


crates/core/action-pipeline/src/estimator.rs
```diff
@@ -4,25 +4,29 @@ use rustc_hash::FxHashMap;
...
     /// How long the actions would have taken to execute outside of moon.
     pub duration: Duration,
 
-    /// Longest duration of each task bucketed by name.
-    pub tasks: FxHashMap<String, Duration>,
+    /// How much less time was spent using moon's pipeline, compared to the baseline.
+    pub gain: Option<Duration>,
 
-    /// How much time was saved using moon's pipeline.
-    pub savings: Option<Duration>,
```
milesj, line 17: split this out into gain/loss since we cant do negative duration

crates/core/action-pipeline/src/estimator.rs
```diff
@@ -35,44 +39,62 @@ impl Estimator {
                 continue;
             };
 
-            if let ActionNode::RunTarget(_, target) = node {
-                let task_id = Target::parse(target).unwrap().task_id;
+            match node {
+                ActionNode::SetupTool(_)
+                | ActionNode::InstallDeps(_)
+                | ActionNode::InstallProjectDeps(_, _) => {
+                    install_duration += *task_duration;
```
milesj, line 46: including install times now, otherwise, the comparison seems much faster. which is wrong since they also install.

PR #641, opened 2023-02-21
fix: Don't include outputs as inputs.

milesj:


milesj:
This is a better one: https://github.com/moonrepo/moon/pull/649

PR #643, opened 2023-02-22
Add support for BSD operating systems

yonas:


milesj:
@yonas Thanks! I'm assuming the entire thing compiles locally now yeah?

With that being said, this will only fix local dev. We don't release to BSD targets yet.

yonas:
> Thanks! I'm assuming the entire thing compiles locally now yeah?

Yes.

> With that being said, this will only fix local dev. We don't release to BSD targets yet.

That's fine. What's holding you back?

milesj:
@yonas Just have to setup all the infra. I'll look into it this weekend and see if I can get it into the release on monday.

milesj:
We extracted proto out into another repo, so I copied over your changes https://github.com/moonrepo/proto/pull/4

PR #695, opened 2023-03-13
fix: Loosen `package.json` parsing to avoid failures.

milesj:
`package.json` files can be written in far too man ways, so this removes all the fields we don't care about.


shellscape:
Awesome. This looks good and thanks for tackling it so quickly.

PR #718, opened 2023-03-18
new: Warn when an input doesn't exist while hashing.

milesj:
Only applies to literal file paths, not globs, and helps to bubble up misconfigurations.


elmpp (commented):
noice

PR #734, opened 2023-03-21
new: Add project constraints and boundary enforcement.

milesj:
This PR implements project constraints, which are rules enforced by the new enforcer crate. It currently has 2 settings:

- The first enforces that projects of a specific type can only depend on other projects of another specific type. Right now it simply enforces that libraries can only be depended on.
- The second enforced depends on relationships based on a project's tags. This gives users control of what can depend on what.


website/docs/config/workspace.mdx
```diff
@@ -98,6 +98,51 @@ projects:
...
+For example, let's enforce that Next.js projects using the `next` tag can only depend on projects
+that support React using the `react` tag.
+
+```yaml title=".moon/workspace.yml" {2,3}
+constraints:
+  tagRelationships:
+    next: ['react']
+```
+
+> Projects with no tags are considered an allowed relationship for easier interoperability.
```
milesj, line 144: Not sure what the best approach is here.

website/docs/config/workspace.mdx
```diff
@@ -98,6 +98,51 @@ projects:
...
+like.
+
+For example, let's enforce that Next.js projects using the `next` tag can only depend on projects
+that support React using the `react` tag.
+
+```yaml title=".moon/workspace.yml" {2,3}
+constraints:
+  tagRelationships:
+    next: ['react']
+```
```
jpoz, line 142: Could use a bit more of an explanation here maybe show where the next and react tags are getting set

website/docs/config/workspace.mdx
jpoz: So anything can depend on anything?

crates/core/enforcer/src/lib.rs
```diff
@@ -0,0 +1,65 @@
+mod errors;
+
+pub use errors::*;
```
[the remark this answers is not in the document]
milesj: It's used by the function and then reexported

PR #736, opened 2023-03-21
fix: Fix pnpm lockfile parsing.

milesj:
Fixes https://github.com/moonrepo/moon/issues/733


crates/node/lang/src/pnpm/dependency_path.rs
```diff
@@ -32,54 +18,71 @@ impl PnpmDependencyPath {
         !path.starts_with('/')
     }
 
-    pub fn parse(path: &str) -> Result<Self, DependencyPathError> {
+    pub fn parse(path: &str) -> Self {
```
milesj, line 21: Based on https://github.com/pnpm/pnpm/blob/main/packages/dependency-path/src/index.ts#L96

crates/node/lang/src/pnpm/dependency_path.rs
```diff
@@ -32,54 +18,71 @@ impl PnpmDependencyPath {
...
                     is_absolute,
                     name,
                     peers_suffix,
                     version: Some(ver),
-                });
+                };
             }
         }
-        if !is_absolute {
-            return Err(DependencyPathError::IsNotAbsolute(path.to_string()));
```
milesj, line 74: I'm removing the error entirely. The error was copied from the JS implementation, but we really don't care if this is wrong, we simply need to proceed.

elmpp:
💯

PR #788, opened 2023-04-17
fix: MOON_COLOR number env not work

### LINKED ISSUE #787: [bug][v1.1.1] `MOON_COLOR=2` env is regarded as bool arg and not work
**Describe the bug**
v1.1.1

**Steps to reproduce**

`MOON_COLOR=2 moon run pkg-a:build`

**Expected behavior**

As expected in the document

**Screenshots**
image

**Environment**

```

```

**Additional context**

SoonIter:
closes #787

### DIFF
--- a/crates/cli/src/app.rs
+++ b/crates/cli/src/app.rs
@@ -587,7 +587,6 @@ pub struct App {
     #[arg(
         long,
         global = true,
-        env = "MOON_COLOR",
         help = "Force colored output for moon"
     )]
     pub color: bool,


milesj:
@SoonIter How does it not work exactly? We have a bunch of code and tests around this: https://github.com/moonrepo/moon/blob/master/crates/cli/src/helpers.rs#L46

milesj:
Copying this fix here so I can get a release out today: https://github.com/moonrepo/moon/pull/789

PR #922, opened 2023-06-19
Fix broken get_local_branch when using older git versions

ilyasotkov:
When running git <2.22.0 (I had 2.14.1), I get

```
moon run build
Error:   × Process git failed with a 1 exit code.
  │ 
  │ git: 'rev-parse --abbrev-ref HEAD' is not a git command. See 'git --help'.
```


milesj:
@ilyasotkov Thanks for the fix. You'll need to run `yarn version:bump:bin patch` for the PR to pass.

PR #1068, opened 2023-09-23
docs: Fix escaping on git-tag-pattern examples

tom-fletcher:
The examples for the `git-tag-pattern` for a `proto` toml plugin need some additional escaping. This is a minor update to the docs to escape these correctly.

(Note - although the default pattern shows how it is represented *internally*, it needs the additional escaping to be used in the toml file).


milesj:
Aeswome thanks.

Issue #1168, opened 2023-11-08
[bug] `Unsupported URL Type "workspace:":`


mctrafik:
I'm new to trying `moon`. I've setup a demo repo with just 2 packages (demo1 and demo2) and I'm referencing one from the other. When I try to run any commands on the second package, I get an error: 

```
 demo2 % moon run demo2:compile
▪▪▪▪ npm install
/Users/user/.proto/tools/npm/10.2.3/shims/npm install --no-audit --no-fund (in workspace)
npm WARN workspaces demo2 in filter set, but no workspace folder present
npm ERR! code EUNSUPPORTEDPROTOCOL
npm ERR! Unsupported URL Type "workspace:": workspace:*

npm ERR! A complete log of this run can be found in: /Users/user/.npm/_logs/2023-11-08T03_51_04_870Z-debug-0.log
Error: process::failed

  × Process ~/.proto/tools/npm/10.2.3/shims/npm failed
  │ with a 1 exit code.
```

From what I can tell, the erroneous lines in `demo2/package.json` are:

```
"dependencies": {
    "demo1": "workspace:*"
  }
```

and are added by `moon`. I don't use other tools, but I understand that this syntax is invalid in `npm`.

I've tried npm v9 and v10. I tried node v18, v20 and v20.9 all with same outcome.

**Steps to reproduce**

toolchain.yml

```yaml
node:
  version: "20.9.0"
  packageManager: "npm"
  npm:
    version: "10.2.3"
```

demo2/moon.yml

```yaml
$schema: "https://moonrepo.dev/schemas/project.json"
dependsOn:
  - "demo1"
language: "typescript"
platform: "node"
```

**Expected behavior**

It should set the path to the `demo1` project that works with `npm`.

**Screenshots**

N/A

**Environment**

```
 System:
    OS: macOS 13.6
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.05 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.12.1 - ~/.nvm/versions/node/v18.12.1/bin/node
    Yarn: 1.13.0 - ~/.yarn/bin/yarn
    npm: 9.6.3 - ~/.nvm/versions/node/v18.12.1/bin/npm
  Managers:
    pip3: 21.2.4 - /usr/bin/pip3
    RubyGems: 3.0.3.1 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
[...]

PR #1173, opened 2023-11-10
fix: Add fallback for unsupported `node.dependencyVersionFormat`.

milesj:
Fixes https://github.com/moonrepo/moon/issues/1168


mctrafik:
Thank you!

PR #1176, opened 2023-11-13
fix: Handful of fixes.

milesj:


tests/repros/1174/.moon/toolchain.yml
```diff
@@ -0,0 +1 @@
+node: {}
```
bryanjswift, line 1: If it matters, this is my `toolchain.yml` file:

```yml
# Configures Node.js within the toolchain. moon manages its own version of Node.js
# instead of relying on a version found on the host machine. This ensures deterministic
# and reproducible builds across any machine.
node:
  # The version to use. Must be a semantic version that includes major, minor, and patch.
  # We suggest using the latest active LTS version: https://nodejs.org/en/about/releases
  version: '18.17.0'

  # The package manager to use when managing dependencies.
  # Accepts "npm" (default), "pnpm", or "yarn".
  packageManager: 'yarn'

  # The version of the package manager (above) to use.
  yarn:
    version: '3.6.3'

  # Add `node.version` as a constraint in the root `package.json` `engines`.
  addEnginesConstraint: true

  # Dedupe dependencies after the lockfile has changed.
  dedupeOnLockfileChange: true

  # Version format to use when syncing dependencies within the project's `package.json`.
  dependencyVersionFormat: 'workspace-caret'

  # Infer and automatically create moon tasks from `package.json` scripts, per project.
  # BEWARE: Tasks and scripts are not 1:1 in functionality, so please refer to the documentation.
  inferTasksFromScripts: false

  # Sync a project's `dependsOn` as dependencies within the project's `package.json`.
  syncProjectWorkspaceDependencies: true

  # Sync `node.version` to a 3rd-party version manager's config file.
  # Accepts "nodenv" (.node-version), "nvm" (.nvmrc), or none.
  # syncVersionManagerConfig: 'nvm'

# Configures how moon integrates with TypeScript.
typescript:
  # When `syncProjectReferences` is enabled and a dependent project reference
  # *does not* have a `tsconfig.json`, automatically create one.
  createMissingConfig: true

  # Name of `tsconfig.json` file in each project root.
  # projectConfigFileName: 'tsconfig.json'

  # Name of `tsconfig.json` file in the workspace root.
  # rootConfigFileName: 'tsconfig.json'

  # Name of the config file in the workspace root that defines shared compiler
  # options for all project reference based config files.
  # rootOptionsConfigFileName: 'tsconfig.options.json'

  # Update a project's `tsconfig.json` to route the `outDir` compiler option
  # to moon's `.moon/cache` directory.
  routeOutDirToCache: false

  # Sync a project's `dependsOn` as project references within the
  # project's `tsconfig.json` and the workspace root `tsconfig.json`.
  syncProjectReferences: false

  # Sync a project's project references as import aliases to the `paths`
  # compiler option in each applicable project.
  syncProjectReferencesToPaths: false
```

tests/repros/1174/chat/moon.yml
```diff
@@ -0,0 +1,6 @@
+tags: [component]
+
+workspace:
+  inheritedTasks:
+    rename:
+      buildComponent: build
```
bryanjswift, line 6: I don't know if the order matters, I have these declared like this:

```yml
workspace:
  inheritedTasks:
    # Exclude the inherited `build` task for this project
    exclude:
      - 'build'
    rename:
      buildComponent: 'build'

tags:
  - 'component'
```

tests/repros/1174/chat/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "chat",
+  "version": "1.0.0",
+  "peerDependencies": {
+    "container": "^1.0.0",
+    "base": "^1.0.0"
+  },
```
bryanjswift, line 7: I also have these dependencies included in `devDependencies` without which editing the type files has odd behavior in vscode:

```suggestion
  "peerDependencies": {
    "container": "^1.0.0",
    "base": "^1.0.0"
  },
  "devDependencies": {
    "container": "workspace:1.0.0",
    "base": "workspace:1.0.0",
  },
```

bryanjswift (commented):
Some comments with additional details about my setup

tests/repros/1174/container/moon.yml
```diff
@@ -0,0 +1 @@
+tags: [component]
```
bryanjswift, line 1: My setup has this:

```suggestion
workspace:
  inheritedTasks:
    # Exclude the inherited `build` task for this project
    exclude:
      - 'build'
    rename:
      buildComponent: 'build'

tags:
  - 'component'
```

tests/repros/1174/base/moon.yml
```diff
@@ -0,0 +1 @@
+tags: [component]
```
bryanjswift, line 1: My setup has this:

```suggestion
workspace:
  inheritedTasks:
    # Exclude the inherited `build` task for this project
    exclude:
      - 'build'
    rename:
      buildComponent: 'build'

tags:
  - 'component'
```

milesj:
@bryanjswift It looks like the `exclude: [build]` is the problem.

This is a bit tricky, let me figure it out.

bryanjswift:
> @bryanjswift It looks like the `exclude: [build]` is the problem.
> 
> Is there another `build` task being inherited that you're trying to filter?

There is, yeah. There is a `build` task that is inherited for all `typescript` projects that I want to ignore for projects with the `component` tag.

milesj:
@bryanjswift Yeah that's tough, since all global tasks are basically in the same hashmap, filtering `deps` doesn't know where it came from.

However, I think `^:buildComponent` may work instead of `^:build`, since we rename it in `deps` too.

PR #1208, opened 2023-11-27
new: Use a `PATH` based approach for task execution.

milesj:
Previously, when we would execute a task, we would attempt to find the real binary for the command on the file system, and execute it directly. For simple tools like bun/node/rust, this was super simple. But for packages within a tool/platform, like node modules and cargo bins, this was extremely complicated and brittle.

To remedy this, we're now using a shell + `PATH` based approach, where we rely on the OS to find the applicable binaries. We just need to ensure all the correct paths are aggregated.


rotu:
I don't use `moon` meaningfully yet, so take my feedback with a heaping tablespoon of salt.

I personally try to keep tasks as shell-agnostic as possible - it's annoying when, e.g. I clone a `npm` project on Windows only to find that the package scripts call `rm -rf` and need to be modified to run! Or to find that the scripts use some bash-specific behavior which doesn't work on Mac (which uses `zsh` by default).

Even on Windows, I don't know if I'd actually want all the stuff (like the many [PowerShell cmdlet abbreviations](https://devblogs.microsoft.com/scripting/table-of-basic-powershell-commands/)) to resolve in a script, unless I opt in specifically.

PR #1258, opened 2024-01-04
Force arch when running in a bun postinstall process

elmpp:
Fixes problem when installing via Bun package manager.

Bun runs its package manager process with the correct process.arch var. This env var influences Moon's arch-specific core dependency.

Bun runs postinstall processes however via node. Node does not have, or does not have set, the same process.arch which results in the postinstall process resolving the wrong core dependency

Bun thankfully sets a useful value at process.npm_config_user_agent which allows a safe edge case solution here

### DIFF
--- a/.yarn/versions/baeeda33.yml
+++ b/.yarn/versions/baeeda33.yml
@@ -1,0 +1,2 @@
+releases:
+  "@moonrepo/cli": patch
--- a/.yarn/versions/baeeda33.yml
+++ b/.yarn/versions/baeeda33.yml
@@ -1,2 +1,9 @@
 releases:
   "@moonrepo/cli": patch
+  "@moonrepo/core-linux-arm64-gnu": patch
+  "@moonrepo/core-linux-arm64-musl": patch
+  "@moonrepo/core-linux-x64-gnu": patch
+  "@moonrepo/core-linux-x64-musl": patch
+  "@moonrepo/core-macos-arm64": patch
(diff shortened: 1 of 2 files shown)


.yarn/versions/baeeda33.yml
```diff
@@ -0,0 +1,2 @@
+releases:
+  "@moonrepo/cli": patch
```
milesj, line 2: need to bump all the core packages too, try `yarn version:bump:bin patch`

elmpp: 👍

packages/cli/postinstall.js
```diff
@@ -13,7 +13,8 @@ const isMacos = process.platform === 'darwin';
 const isWindows = process.platform === 'win32';
 
 const platform = isWindows ? 'windows' : isMacos ? 'macos' : process.platform;
-const parts = [platform, process.arch];
+const arch = process.env['npm_config_user_agent'].match(/^bun.*arm64$/) ? 'arm64' : process.arch; // https://github.com/moonrepo/moon/issues/1103
```
milesj, line 16: @elmpp Is `npm_config_user_agent` always set, even for npm/pnpm/yarn? Maybe we need a property existence check?

elmpp: Added an old-style conditional 👍

milesj:
awesome thanks!

elmpp:
Thank you 🙏

PR #1276, opened 2024-01-18
Allow dependency skipping in task config when missing

maastrich:
Allow tasks to be skipped if they are missing.

```yaml
// .moon/tasks/typescript.yml
tasks:
  typecheck:
    platform: node
    command: tsc --build
    deps:
      - title: "codegen"
        skipIfMissing: true
```

```yaml
// .moon/tasks/tag-codegen.yml
tasks:
  codegen:
    platform: node
    command: do things
```

This won't ve necessary anymore:

```yaml
// .moon/tasks.yml
tasks:
  codegen:
    command: noop
```


nextgen/config/src/project/task_config.rs
```diff
@@ -97,6 +97,8 @@ cacheable!(
         pub env: FxHashMap<String, String>,
 
         pub target: Target,
+
+        pub skip_if_missing: Option<bool>,
```
milesj, line 101: Let's just call it `optional`. Not sure skip if missing would make sense to end users.

nextgen/project-expander/src/tasks_expander.rs
```diff
@@ -159,11 +164,14 @@ impl<'graph, 'query> TasksExpander<'graph, 'query> {
...
                         // Avoid circular references
                     } else {
-                        check_and_push_dep(project, dep, false)?;
+                        check_and_push_dep(project, dep, dep.skip_if_missing.unwrap_or(false))?;
                     }
                 }
                 // id:task
                 TargetScope::Project(project_locator) => {
+                    if dep.skip_if_missing.is_some() {
+                        // log a message to the user to let them know that this is effectless
```
milesj, line 173: I think on line 185 you'll need to handle the error case, because it won't exist sometimes.

milesj (changes requested):
Great start. A few other things:

- Can you change the base to `develop-1.20` instead of `master`.
- Gonna need a few tests!

nextgen/config/src/project/task_config.rs
maastrich: agreed: https://github.com/moonrepo/moon/pull/1276/commits/05b29703d4f7da515f9b0af8391f0ad7d6ed31fb

nextgen/project-expander/src/tasks_expander.rs
maastrich: Not sure, I just wanted to log a message (not changing this case) 🤔  (line 175 and 189 are left unchanged)
Except if we want to allow skipping project dependencies 🤔

nextgen/project-expander/tests/tasks_expander_test.rs
```diff
@@ -809,6 +809,53 @@ mod tasks_expander {
...
+                let context =
+                    create_context_with_query(&project, sandbox.path(), |i| query.filtered(i));
+                TasksExpander::new(&context).expand_deps(&mut task).unwrap();
+
+                assert_eq!(task.deps, vec![]);
+            }
+        }
+
+        #[test]
+        fn not_skip_missing_targets() {
```
milesj: looks like this isnt in the module block on line 834, can you move it up

nextgen/project-expander/tests/tasks_expander_test.rs
```diff
@@ -809,6 +809,53 @@ mod tasks_expander {
                     ]
                 );
             }
+
+            #[test]
+            fn skip_missing_targets() {
```
milesj: can you also add a test where it doesnt skip if the target does exist, and optional = true

milesj (approved):

Issue #1281, opened 2024-01-23
[bug] Moon main thread panicked while using git version 2.43.0.vfs.0.0


huandzh:
**Describe the bug**

moon main thread panicked with the installed Git version "2.43.0.vfs.0.0".

```shell
moon run app:build
Error:   × Main thread panicked.
  ├─▶ at nextgen\vcs\src\git.rs:639:36
  ╰─▶ called `Result::unwrap()` on an `Err` value: Error("unexpected character '.' after patch version number")
  help: set the `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

**Steps to reproduce**

1. has a git installation with version 2.43.0.vfs.0.0
2. run any task from the root of the moon repo

**Expected behavior**

moon can run tasks

**Environment**

```
  System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz
    Memory: 3.24 GB / 15.93 GB
  Binaries:
    Node: 21.5.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.21 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.2.4 - C:\Program Files\nodejs\npm.CMD
  Managers:
    Cargo: 1.70.0 - ~\.cargo\bin\cargo.EXE
    pip3: 23.0.1 - C:\Python311\Scripts\pip3.EXE
  Utilities:
    CMake: 3.23.0
    Git: 2.43.0.
    Clang: 16.0.2 - C:\Program Files\LLVM\bin\clang.EXE
    Curl: 8.0.1 - C:\Windows\system32\curl.EXE
  Virtualization:
    Docker: 24.0.6 - C:\Program Files\Docker\Docker\resources\bin\docker.EXE
  IDEs:
    VSCode: 1.85.1 - C:\Users\p\AppData\Local\Programs\Microsoft VS Code\bin\code.CMD
    Visual Studio: 17.8.34408.163 (Visual Studio Community 2022)
  Languages:
    Bash: 5.1.4 - C:\Windows\system32\bash.EXE
    Python: 3.11.0
    Rust: 1.70.0
  Browsers:
    Edge: Spartan (44.19041.1266.0), Chromium (120.0.2210.144), ChromiumDev (122.0.2348.0)
    Internet Explorer: 11.0.19041.1566
```

**Additional context**

```shell
git --version
git version 2.43.0.vfs.0.0
```

[...]

PR #1282, opened 2024-01-23
fix:Fix version cleaning for VFSForGit windows (#1281).

huandzh:
VFSForGit has the version '2.43.0.vfs.0.0', which has raised issue #1281. 

The proposed fix includes a rewritten regex and an added assert.


milesj:
@huandzh Thanks for the PR, the change looks good.

So forks are broken right now and we fixed it, but we fixed it in the develop-1.20 branch. Think you could rebase off that and change the base branch in this PR?

huandzh:
You're welcome. I'm glad you like the change. The commit has been rebased as suggested.

milesj:
Thanks! Will get out in a few days.

PR #1290, opened 2024-01-26
fix(website): comparison table is significantly mistaken about nx

JamesHenry:
Hi @milesj,

Congrats on all the great work with moon and proto!

I was a bit dismayed to see how significantly mistaken the website's comparison table is about Nx, so I have made the necessary changes here for your review.

Please let me know if anything is unclear or you want to discuss further.

Cheers!

### DIFF
--- a/website/src/components/ComparisonTable.tsx
+++ b/website/src/components/ComparisonTable.tsx
@@ -19,7 +19,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Written in',
 		support: {
 			moon: 'Rust',
-			nx: 'Node.js',
+			nx: 'Rust for speed & TypeScript for extensibility',
 			turborepo: 'Go (being rewritten in Rust)',
 		},
 	},
@@ -43,7 +43,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Repo / folder structure',
 		support: {
 			moon: 'loose',
-			nx: 'strict',
+			nx: 'loose',
 			turborepo: 'loose',
 		},
 	},
@@ -59,6 +59,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Supports dependencies inherited by all tasks',
 		support: {
 			moon: [SUPPORTED, 'via `implicitDeps`'],
+			nx: [SUPPORTED, 'via `targetDefaults`'],
 		},
 	},
 	{
@@ -73,6 +74,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Supports tasks inherited by all projects',
 		support: {
 			moon: SUPPORTED,
+			nx: [SUPPORTED, 'via `plugins`'],
 		},
 	},
 	{
@@ -98,7 +100,7 @@ const toolchainRows: Comparison[] = [
 		feature: 'Supported languages',
 		support: {
 			moon: 'Bash, Batch, Bun, Deno, Go, Node, JavaScript, TypeScript, PHP, Ruby, Rust, Python',
-			nx: 'JavaScript, TypeScript',
+			nx: 'All languages via plugins. OOTB TS/JS, existing plugins for Rust, Go, Dotnet and more',
 			turborepo: 'JavaScript, TypeScript',
 		},
 	},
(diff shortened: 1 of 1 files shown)


website/src/components/ComparisonTable.tsx
```diff
@@ -43,7 +43,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Repo / folder structure',
 		support: {
 			moon: 'loose',
-			nx: 'strict',
+			nx: 'loose',
```
JamesHenry, line 46: This row now accurately provides no distinction between the tools so may ultimately want to remove it. Nx does not care about your folder structure, you can organize things however you want.

JamesHenry:
It is not in the diff of the PR, but it was not clear to me what is meant by this entry:

```ts
{
		feature: 'Supports pipes, redirects, etc',
		support: {
			moon: [PARTIALLY_SUPPORTED, 'encapsulated in a file'],
			nx: [PARTIALLY_SUPPORTED, 'within the executor or script'],
			turborepo: [PARTIALLY_SUPPORTED, 'within the script'],
		},
	},
```

Can you clarify what is meant by "supporting pipes, redirects, etc" in Nx is "within the executor or script" whereas moon is "encapsulated in a file"?

Maybe you could give an example of what you mean?

website/src/components/ComparisonTable.tsx
```diff
@@ -98,7 +100,7 @@ const toolchainRows: Comparison[] = [
 		feature: 'Supported languages',
 		support: {
 			moon: 'Bash, Batch, Bun, Deno, Go, Node, JavaScript, TypeScript, PHP, Ruby, Rust, Python',
-			nx: 'JavaScript, TypeScript',
+			nx: 'All languages via plugins. OOTB TS/JS, existing plugins for Rust, Go, Dotnet and more',
```
milesj, line 103: @JamesHenry This section is mostly about an integrated toolchain. A plugin system supporting other languages really isn't the same thing.

We may just need another section for this, but I can do that after merging.

website/src/components/ComparisonTable.tsx
```diff
@@ -19,7 +19,7 @@ const workspaceRows: Comparison[] = [
 		feature: 'Written in',
 		support: {
 			moon: 'Rust',
-			nx: 'Node.js',
+			nx: 'Rust for speed & TypeScript for extensibility',
```
milesj, line 22: @JamesHenry Do you have more on this? It seems like Nx only uses Rust for a very small portion, so this sentence seems disingenuous.

milesj (commented):
Thanks for this. It's been like a good year since this has been updated. Just a few questions.

website/src/components/ComparisonTable.tsx
JamesHenry: It's your website, so you are free to structure however you wish. But it is incorrect to imply that Nx only supports JavaScript and TypeScript.

Nx is architected in a similar way to VSCode or other IDEs, we have an entirely agnostic core that deals with project graph construction, task running and caching etc, and then everything else is applied via plugins. The Nx Team provides many such plugins so that things don't need to be reinvented, just like the VSCode team does, but we have a ton from the community as well: https://nx.dev/plugin-registry

This plugability very much includes programming languages, and we have plenty of precedent of mixed language monorepos powered by Nx.

website/src/components/ComparisonTable.tsx
JamesHenry: This is about importance vs quantity.

Rust is very important to Nx but we are diligent about benchmarking, and only look to write things in Rust when it provides an objective, real-world speed benefit over TypeScript. In practice this is less often than people assume.

As per my previous comment Nx is architected in such a way that it has a language and framework agnostic core, and then receives that context specific knowledge via plugins.

It is only the Nx Core that is in scope for Rust, because having plugins in Rust has a high cost to the community. Writing plugins in TS/JS has a lower barrier to entry than Rust and is one of the reasons why it has proven to be so successful and why we receive so many community PRs to our first-party plugins. We have a lot of first-party plugins in our monorepo, so basing your comment around the number of files in one programming language vs another then it will appear heavily diluted.

Nx uses Rust for critical operations such as:
- project graph construction
- file system operations, including file watching
- globbing
- hashing
- new task runner which will be the default in the next version

These are the most important things that Nx does, and they are written in Rust. It is not disingenuous to say that Nx is written in Rust for speed and TypeScript for extensibility.

Hopefully that clarifies things!

website/src/components/ComparisonTable.tsx
milesj: I don't disagree, but in moon there's a distinction between toolchain languages, and languages supported via plugins. It's not really reflected right now, so I'll update it after the fact.

Issue #1284, opened 2024-01-23
[feature] Use `git-bash` on Windows


shellscape:
**Is your feature request related to a problem? Please describe.**

I'm attempting to run some tasks on Ubuntu and Windows (via Github Actions) and things are not going well. Many commands that use *nix staples like `rm -rf` simply don't work on Windows. Since moon uses powershell by default (according to https://moonrepo.dev/docs/config/project#shell) on Windows, we're limited in what we can do. And I'd rather not use a node module for simple tasks like copying files and removing directories, or redirect out to a tool that can use the correct shell for the need (like pnpm/npm scripts).

**Describe the solution you'd like**

I'd argue that Git Bash is widely available on machines that are likely to be running moon, and is a handy tool for cross-platform work. Tools like pnpm have considered this and have guidance https://pnpm.io/cli/run#script-shell

Being able to specify which shell to use based on bath, or by name if moon wants to try and locate it, would be incredible. Specifically for using git-bash in windows environments.

**Describe alternatives you've considered**

The sad panda workaround path is to perform these tasks in a pnpm package.json script and have moon execute pnpm to run them in the git-bash shell. 

**Additional context**

milesj:
Added `unixShell` and `windowsShell` in v1.21, can you give those a shot

shellscape:
Will be able to give that a spin towards the end of the week, got a pile in front of me atm

milesj:
Gonna close this. Feel free to re-open if there are issues.

shellscape:
Gah. I'm the worst, sorry about that. We're good to go.

PR #1303, opened 2024-02-01
new: Allow Unix and Windows shells to be customizeable.

milesj:
Fixes https://github.com/moonrepo/moon/issues/1284

cc @shellscape


shellscape:
Nice..Thanks for getting this in

Issue #1329, opened 2024-02-10
[feature] affectedFiles should pass in all input files when there are no affected files instead of `.`


affanshahid:
**Is your feature request related to a problem? Please describe.**

I really like how the `--affected` option behaves especially when working with pre-commit hooks for formatting/linting tasks. However when you want to format the entire project (say when you wipe the cache or another such scenario), there is an issue. By default `.` is passed in when there are no affected files regardless of what the input is, this makes it impossible to filter out certain files from the formatting. I could simply use `@files()` or another token in the command args directly to manually exclude files but then you lose the benefits of `--affected` with `affectedFiles: true`.

**Describe the solution you'd like**

When there are no affected files, moon should pass in ALL the files matched by the inputs instead of `.`. Moon should resolve the input globs as it normally does for affected and pass in the total result instead of just `.`.

**Describe alternatives you've considered**

The current solution requires using the underlying tools ignore options (i.e .prettierignore) but not all formatting tools might support something like this and it adds yet another configuration file to the repo. I'd like this to be configurable through moon since it already knows which files my task operates on.

**Additional context**

milesj:
This seems reasonable, I can look into it.

milesj:
Landed in 1.22

milesj:
Reopening as this seems to cause issues with other tasks: https://github.com/moonrepo/moon/issues/1353

We may have to wrap this in a setting and be opt-in.

milesj:
Added an option `affectedPassInputs` for this, to avoid other issues.

PR #1331, opened 2024-02-11
feat: default to all input files for affected tasks

affanshahid:
Just sending a PR for what I had to do locally to "fix" https://github.com/moonrepo/moon/issues/1329. Feel free to close this if someone else is already working on this or if my approach isn't ideal.

Wasn't sure where to put the "walking" code, one option was to create a fake FileGroup to re-use the logic from there but that felt hacky, so added a method to Task.


nextgen/task/src/task.rs
```diff
@@ -159,6 +166,43 @@ impl Task {
...
+    pub fn get_input_files<S: AsRef<str>>(
+        &self,
+        workspace_root: &Path,
+        project_source: S,
+    ) -> miette::Result<Vec<ProjectRelativePathBuf>> {
+        let mut list = vec![];
+        let project_source = project_source.as_ref();
+
+        for path in &self.input_files {
+            list.push(path.strip_prefix(project_source).unwrap().to_owned());
```
milesj, line 180: I think this should return the paths relative from the workspace root.

The "relative from the project root" is only required for the affected logic, so I would move that part over there.

milesj:
New features also go into the develop branch, can you change the base to `develop-1.22`.

affanshahid:
I have made the requested changes

milesj:
@affanshahid Implementation looks good, just needs some tests.

affanshahid:
@milesj I've added tests, lmk if something seems off. However one thing I realised is that if the the set of input files is empty then the underlying command will be passed no arguments. This might break things for people expected a `.` in those scenarios, what do you think? Maybe add a "." arg as a fallback to be deprecated in a future breaking release?

milesj:
@affanshahid Good point. Maybe in `runner.rs`, if files list is still empty, default to `.`.

affanshahid:
@milesj I've added the fallback along with associated tests and a deprecation warning

milesj:
@affanshahid changes looked good, just need to rebase the develop branch since changes landed there

affanshahid:
@milesj Just did, also changed tests for deno, node, bun, system which expected a dot when no files are expected

affanshahid:
@milesj I've rebased and updated the hashses

milesj:
@affanshahid Awesome, thanks you!

PR #1410, opened 2024-04-03
docs: Add Hugo standard and extended for proto.

z0rrn:
Hugo has two version/editions thus I provide a plugin for both.

Thanks! Frederik Zorn


z0rrn:
I ran the test locally now and everything worked. Please approve again.

Thanks

milesj:
Thanks!

PR #1416, opened 2024-04-06
feat: add mutex option to task configuration

### LINKED ISSUE #1415: [feature] being able to set a "mutex" option for tasks
## Is your feature request related to a problem? Please describe:

I have some "local" tasks that are defined at a global level and are executed for all the sub-projects.

Some of these tasks require having exclusive access over certain filesystem resources while they are running (for example, they might need exclusive access over `./.git/index.lock`, but that's not the only case).

I tried to specify the "exclusive resource" path in the `inputs` and `outputs` sections, believing that parallelisation would be reduced if both tasks tried to read/write the same resources... but Moon doesn't seem to use that information to infer that these tasks can run at the same time.

It could be as well that because my "exclusive resource" was in `./.git/`, Moon didn't even try to factor it into its graph computations (I guess it is an ignored file). 

### Solution 1 (not my preferred, just to be more exhaustive)

I think it would be reasonable that Moon takes into consideration the fact that some tasks share the same inputs and outputs to avoid running them at the same time.

I suppose that, under "simple" cases (inputs and output are different from each other), Moon can use this information to refine its dependencies graph beyond what can be obtained from explicit dependencies and dependencies extracted from `package.json` or similar files...

But I suspect that some cases are more complicated, like having multiple tasks with the same `outputs` and different inputs, which gives us no clue as to "who" should have precedence over that file... Although it certainly tells us that, independently of their order, they shouldn't run at the same time because they are generating side effects over the same resource.

Alternatively, if using `outputs` to infer incompatibilities implies having to change too much in Moon's codebase, we could introduce some new field, something like `exclusiveResources`:
```yaml
tasks:
  my.special.task:
    command: 'do something'
    exclusiveResources: # Refers to filesystem resources
      - 'exclusive_resource.lock'
```

The problem with this solution is that it wouldn't be able to consider "resources" outside the immediate scope of Moon (for example, I suspect that "ignored files" wouldn't be factored into the graph computations, as mentioned in the previous section).

### Solution 2 (my preferred)

It would help a great deal being able to add an attribute to some tasks to indicate if they can be run in parallel with their "twins" running in other projects, or, more generally, to indicate if they are "incompatible" with a list of other task references (incompatible to be executed at the same time as the others).

What I think is ideal, with a "virtual mutex", it would make it "incompatible" for parallel execution with any other task sharing the same `mutex` value.
```yaml
tasks:
  my.special.task:
    command: 'do something'
    options:
      # this is an "abstract mutex", not tied to the filesystem, so we
      # don't risk problems related to ignored files, and also allows us
      # to "protect" remote resources that can't be modelled by Moon.
      mutex: 'resource_1'
```

castarco:
## Description

This PR introduces a new `mutex` option to task options. This new option is intended to allow telling Moon when two or more tasks can't be executed at the same time.

Example:

```yaml
tasks:
  task_a:
    command: 'do something'
    options:
      # this is an "abstract mutex", not tied to the filesystem, so we
      # don't risk problems related to ignored files, and also allows us
      # to "protect" remote resources that can't be modelled by Moon.
      mutex: 'resource_1'
  task_b:
    command: 'do something else'
    options:
      mutex: 'resource_1' # task_a and task_b won't run at the same time
```

Closes #1415 .

## Pending tasks
- [X] Complete PR description
- [x] Remove some `unwrap` calls
- [ ] Add new automated tests
- [x] Exhaustive manual testing (I already did some testing with good results, but I didn't stress the tool too much yet).

### DIFF
--- a/crates/core/action-context/src/lib.rs
+++ b/crates/core/action-context/src/lib.rs
@@ -3,7 +3,7 @@ use moon_common::path::WorkspaceRelativePathBuf;
 use moon_target::{Target, TargetLocator};
 use rustc_hash::{FxHashMap, FxHashSet};
 use serde::{Deserialize, Serialize};
-use std::path::PathBuf;
+use std::{path::PathBuf, sync::Arc};
 
 #[derive(ValueEnum, Clone, Debug, Deserialize, Serialize)]
 pub enum ProfileType {
@@ -26,12 +26,58 @@ impl TargetState {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct TaskNamedMutexes {
+    mutexes: Arc<std::sync::Mutex<FxHashMap<String, Arc<tokio::sync::Mutex<()>>>>>,
+}
+
+impl<'de> Deserialize<'de> for TaskNamedMutexes {
+    fn deserialize<D>(_: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        // We don't care about unserializing this, but we need to satisfy the
+        // trait requirements.
+        Ok(TaskNamedMutexes::new())
+    }
+}
+
+impl Serialize for TaskNamedMutexes {
+    fn serialize<S>(&self, _serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        // We don't care about serializing this, but we need to satisfy the
+        // trait requirements.
+        FxHashMap::<String, ()>::default().serialize(_serializer)
+    }
+}
+
+impl TaskNamedMutexes {
+    fn new() -> Self {
+        TaskNamedMutexes {
+            mutexes: Arc::new(std::sync::Mutex::new(FxHashMap::default())),
+        }
+    }
+
+    pub fn get(&self, name: &str) -> Arc<tokio::sync::Mutex<()>> {
+        // TODO: Check how to remove that `unwrap`
+        let mut mutexes = self.mutexes.lock().unwrap();
+        if !mutexes.contains_key(name) {
+            mutexes.insert(name.to_string(), Arc::new(tokio::sync::Mutex::new(())));
+        }
+        mutexes.get(name).unwrap().clone()
+    }
+}
+
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ActionContext {
     pub affected_only: bool,
 
     pub initial_targets: FxHashSet<TargetLocator>,
+
+    pub named_mutexes: TaskNamedMutexes,
 
     pub passthrough_args: Vec<String>,
 
--- a/crates/core/action-context/src/lib.rs
+++ b/crates/core/action-context/src/lib.rs
@@ -31,35 +31,7 @@ pub struct TaskNamedMutexes {
     mutexes: Arc<std::sync::Mutex<FxHashMap<String, Arc<tokio::sync::Mutex<()>>>>>,
 }
 
-impl<'de> Deserialize<'de> for TaskNamedMutexes {
-    fn deserialize<D>(_: D) -> Result<Self, D::Error>
-    where
-        D: serde::Deserializer<'de>,
-    {
-        // We don't care about unserializing this, but we need to satisfy the
-        // trait requirements.
-        Ok(TaskNamedMutexes::new())
-    }
-}
-
-impl Serialize for TaskNamedMutexes {
-    fn serialize<S>(&self, _serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: serde::Serializer,
-    {
-        // We don't care about serializing this, but we need to satisfy the
-        // trait requirements.
-        FxHashMap::<String, ()>::default().serialize(_serializer)
-    }
-}
-
 impl TaskNamedMutexes {
-    fn new() -> Self {
-        TaskNamedMutexes {
-            mutexes: Arc::new(std::sync::Mutex::new(FxHashMap::default())),
-        }
-    }
-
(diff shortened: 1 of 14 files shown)


crates/core/action-context/src/lib.rs
```diff
@@ -26,13 +26,59 @@ impl TargetState {
...
+}
+
 #[derive(Clone, Debug, Default, Deserialize, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct ActionContext {
     pub affected_only: bool,
 
     pub initial_targets: FxHashSet<TargetLocator>,
 
+    pub named_mutexes: TaskNamedMutexes,
```
milesj, line 80: Probably wrap with `#[serde(skip)]` instead of the impls above?

It feels weird putting this on the context, but there's no other good place at the moment.

crates/core/action-context/src/lib.rs
```diff
@@ -26,13 +26,59 @@ impl TargetState {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct TaskNamedMutexes {
+    mutexes: Arc<std::sync::Mutex<FxHashMap<String, Arc<tokio::sync::Mutex<()>>>>>,
```
milesj, line 31: Mutexes within a mutex may be much. Maybe dashmap will work better here? https://crates.io/crates/dashmap

crates/core/action-pipeline/src/actions/run_task.rs
```diff
@@ -95,7 +95,23 @@ pub async fn run_task(
     }
 
     let attempts_result = if is_cache_enabled {
-        let context = context.read().await;
+        if let Some(mutex_name) = &task.options.mutex {
```
milesj, line 98: This feels... messy? But honestly, not sure there's another way :P

crates/core/action-context/src/lib.rs
castarco: Ah, I didn't know about `#[serde(skip)]`, that makes much more sense than the hack I implemented.

It's been like 2 years since my last meaningful coding sessions in Rust 😅 .

crates/core/action-context/src/lib.rs
castarco: Dashmap sounds pretty good, I admit my implementation is pretty naive. I'll try it to see if it behaves well 😃 .

crates/core/action-pipeline/src/actions/run_task.rs
castarco: Yep, it is. I'll give it some extra thought. I don't promise anything, but I'd like it to be a bit cleaner, if possible.

crates/core/action-context/src/lib.rs
castarco: Done :) .

crates/core/action-pipeline/src/actions/run_task.rs
castarco: @milesj 

I reworked the execution flow to avoid having nested `if-else` control structures.

This way the control flow is "simpler", but it involves 2 extra allocations and 2 extra indirections (for `Arc::new` and `RwLock::new`) that are necessary to match the types for the cases when `is_cache_enabled` is `true` and `false`.

One possibility can be to go back to a "more branched" approach, but extracting the repeated logic into its own function. For now I'll leave it as it is so we can discuss it.

crates/core/action-pipeline/src/actions/run_task.rs
```diff
@@ -94,18 +94,34 @@ pub async fn run_task(
...
+    let attempts_result = {
+        let _ctx: Arc<RwLock<ActionContext>>;
+        let ctx = if is_cache_enabled {
+            &context
+        } else {
+            // Concurrent long-running tasks will cause a deadlock, as some threads will
+            // attempt to write to context while others are reading from it, and long-running
+            // tasks may never release the lock. Unfortuantely we have to clone here to work
+            // around it, so revisit in the future.
+            _ctx = Arc::new(RwLock::new(context.read().await.clone()));
```
castarco, line 106: This is the part that I don't like.

Although creating an `Arc<RwLock<*>>` instance does help to match the types of both branches, it involves 2 extra allocations, and 2 extra indirections once we access the value a few lines below.

milesj:
I think the code looks fine. Just need to fix the CI builds.

I'm also unsure of how to unit test this properly :P

castarco:
> I'm also unsure of how to unit test this properly :P

I'll think of something before the week ends, at least some coverage would be good.

milesj:
@castarco Planning to release 1.24 on wednesday. Think you could get this done by then?

castarco:
Hi @milesj , I'm sorry for the delay on my response. I'm spread too thin these days. I'll try, but I can't promise anything 😞 .

castarco:
btw @milesj , in my local environment, there are some tests that seem a bit flaky, they pass sometimes, and they fail sometimes. They are mostly related to Yarn, Bun, and NPM.

milesj:
@castarco Yeah some tests are flaky, can just ignore them.

Just need to fix the formatting on the docs pages and this is good.

website/tsconfig.json
```diff
@@ -1,13 +1,18 @@
...
     "emitDeclarationOnly": true,
     "noEmit": false,
     "outDir": "../.moon/cache/types/website",
     "verbatimModuleSyntax": false,
     "baseUrl": ".",
     "paths": {
-      "@site/*": ["./*"]
+      "@site/*": [
+        "./*"
+      ]
```
castarco, line 15: I didn't want to change this, but prettier was quite adamant that it didn't like its previous state, so I wanted to minimize the chances that the CI pipeline would fail.

milesj:
Awesome stuff, thanks for spiking this!

PR #1434, opened 2024-04-17
Add Shebang to packages/cli/moon to Resolve Exec Format Error on Linux

reoring:
After cloning the moonrepo repository, I noticed that the contents of `packages/cli/moon` are in plain text and do not include a shebang line. This absence leads to an "Exec format error (os error 8)" when trying to execute the script on Linux, making it difficult to understand the error. 

However, the script merely outputs text when run on macOS. To address this issue and clarify the script's purpose across different operating systems, I have added a shebang line to the script.

### DIFF
--- a/packages/cli/moon
+++ b/packages/cli/moon
@@ -1,3 +1,8 @@
 #!/bin/sh
 
+# This script is a placeholder used by package managers for linking the core binary.
+# To run the moon tool directly from the repository, use the following command:
+# cargo run
+
 echo "This file is necessary for package managers to link the core binary!"
+echo "If you want to run moon from the repo, please use: cargo run"
--- a/packages/cli/moon
+++ b/packages/cli/moon
@@ -1,1 +1,3 @@
-This file is necessary for package managers to link the core binary!
+#!/bin/sh
+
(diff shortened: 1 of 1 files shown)


milesj:
@reoring What are you trying to do exactly?

This file gets replaced by package managers when the package is installed, but should never be executed by end users. If you're trying to run `moon` in the repo itself, we suggest using `cargo run`.

reoring:
Hi @milesj,

Thank you for your prompt response and for clarifying how the script is handled during installation. Based on your advice, I understand that running cargo run is the recommended approach for developers looking to run moon directly from the repository.

Considering this, I propose adding a note or message in directly within the script. This would serve as a helpful guide for new developers, ensuring they use cargo run and understand the workflow more quickly. It could potentially prevent any confusion for those who might try to execute the script directly.

I believe this small addition could significantly enhance the onboarding experience for new contributors. Let me know what you think!

Best regards,

milesj:
@reoring Change looks good, just need to update `scripts/checks/cliBinary.sh`.

You can ignore the version check error.

milesj:
Gonna close this since this PR actually fixes this indirectly https://github.com/moonrepo/moon/pull/1472

PR #1747, opened 2024-12-05
fix: Improve CI job handling.

t3ned:
This PR aims to make some small improvements to moon's CI command based on production experience.

- [x] FIX: The "Job total" log prints `job_index` instead of `job_total` 
- [x] FIX: Uneven weight on the last job index's batched targets 
      (i.e. 10 total jobs, 45 total targets - the last job will have 9 targets instead of 4)
- [ ] ~FEAT: Discuss (& implement if in agreement) whether targets should be distributed based on the action graph~

### DIFF
--- a/.yarn/versions/15d6e39f.yml
+++ b/.yarn/versions/15d6e39f.yml
@@ -1,0 +1,9 @@
+releases:
+  "@moonrepo/cli": patch
+  "@moonrepo/core-linux-arm64-gnu": patch
+  "@moonrepo/core-linux-arm64-musl": patch
+  "@moonrepo/core-linux-x64-gnu": patch
+  "@moonrepo/core-linux-x64-musl": patch
+  "@moonrepo/core-macos-arm64": patch
+  "@moonrepo/core-macos-x64": patch
+  "@moonrepo/core-windows-x64-msvc": patch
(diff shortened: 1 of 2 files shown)


t3ned:
It's possible for parallel CI jobs to run no targets despite an even distribution of actionable targets because the distribution is across *potential* targets.

My proposal is to distribute only the targets that would run in the pipeline instead of all the potential targets. 

And perhaps a bit more complex, but also use the action graph to actually determine an evenly distributed number of actionable targets based on dependencies. For example, imagine we have 4 actionable targets, namely `api:build`, `api:lint`, `api:test`, `api:format`, where `api:test` depends on `api:build`. If we have 2 total parallel jobs (job 0 and job 1), we can run the two dependants (`api:build` and `api:test`) in job 0, and the remaining two (`api:lint` and `api:format`) in job 1. The problem here is now some cases (maybe most depending on the project), we can't evenly distribute the targets, i.e. if there's 3 dependants out of a total 3 actionable targets, and the same 2 jobs, now job 1 has no targets to run because job 0 is running the dependants.

Any thoughts and discussion around this is appreciated!

milesj:
@t3ned This looks good for now. Just need to run `just bump patch`.

> My proposal is to distribute only the targets that would run in the pipeline instead of all the potential targets.

Yeah that's the future plan. When this was original written, we didn't have access to that information, but we should be able to at this point.

As for the other distribution, I tend to agree, but this wouldn't really be feasible without remote caching, as the dependencies of a task will always need to run in the same job as the owning task.

PR #1748, opened 2024-12-07
fix: Some Python improvements.

milesj:


milesj:
@harlequin Some changes based on feedback in Discord. Thoughts?

harlequin:
@milesj have had a look to the changes and seems everything ok. Would go for it.

Issue #1761, opened 2024-12-26
[bug] pre-commit hooks don't propagate exit code in powershell


chadrik:
**Describe the bug**

The autogenerated `pre-commit.ps1` script does not return non-zero when the configured commands return non-zero, and as a result they will not halt a commit from proceeding on error. 

**Steps to reproduce**

Using Powershell on Windows, do the following:

1. Set the pre-commit script to an executable (not another powershell script) that returns non-zero such as `node -e "process.exit(1)"` or `python -c "import sys;sys.exit(1)"`:
```yml
vcs:
  manager: 'git'
  hooks:
    pre-commit:
      - 'python -c "import sys;sys.exit(1)"'
  ```

2. Use `moon sync hooks` to create the `pre-commit.ps1` script.  The result looks like this:

```powershell
#!/usr/bin/env pwsh
$ErrorActionPreference = 'Stop'

# Automatically generated by moon. DO NOT MODIFY!
# https://moonrepo.dev/docs/guides/vcs-hooks

python -c "import sys;sys.exit(1)"
```

3.  Open a powershell console and execute the script: 

```
pwsh .moon\hooks\pre-commit.ps1
```

4.  Print the exit code from the powershell console:

```
Write-Host "exit code $($LASTEXITCODE)"
```

The expectation is that `$ErrorActionPreference = 'Stop'` will cause the script to abort and return non-zero as with `set -e` in Linux, but `$ErrorActionPreference = 'Stop'` only applies to powershell cmlets, not "native" commands (i.e. subproceses), as explained [here](https://stackoverflow.com/a/9949909).

**Expected behavior**

`pre-commit.ps1` should return non-zero when one of its commands fails, so that git commit will be halted. 

In Powershell 7.4 there is [an option](https://learn.microsoft.com/en-us/powershell/scripting/learn/experimental-features?view=powershell-7.4#psnativecommanderroractionpreference) called `$PSNativeCommandUseErrorActionPreference` which, when set to `$true`, will cause the `$ErrorActionPreference = 'Stop'` to apply to native commands, and should produce the expected behavior.  However, 7.4 is still relatively new, so it would be best if moon supported older versions of powershell.   I tested `$PSNativeCommandUseErrorActionPreference` in powershell 7.2 and it has no effect. 

A solution that works for older versions of powershell requires an explicit check of the exit code after each command:

```powershell
#!/usr/bin/env pwsh
$ErrorActionPreference = 'Stop'

[...]

PR #1764, opened 2024-12-28
fix: Fix pwsh git hook exit codes.

milesj:
Fixes https://github.com/moonrepo/moon/issues/1761


chadrik:
Thanks for the quick fix!

PR #1773, opened 2025-01-01
new: Add `moon run --eager` to disable fail-fast behavior and execute as many tasks as possible

### LINKED ISSUE #1770: [feature] Provide an option to disable fail-fast behavior and process as much of the task graph as possible
Thanks for your great project.   I'd like to learn to rust, so if any of the issues I've submitted could be good beginner tasks, let me know. I was halfway to fixing the powershell bug when you submitted your PR 😄 

Currently I use [pre-commit](https://pre-commit.com/) to manage and run my git hooks, but I hate that it requires me to maintain a configuration file with a bunch of information that is redundant with settings owned by moon.  Moon is very close to being a complete replacement for pre-commit, which would give me a single source of truth and a more elegant solution, but it's missing a few features.

Here are the required features (check boxes indicate whether moon satisfies the requirement):

- [x] Install necessary toolchains to run tasks
- [x] Specify file patterns for when to trigger a task (solved with `inputs`)
- [x] Operate on staged files (solved with `--affected --status=staged`)
- [x] Pass modified input files to the task  (solved with `affectedFiles`)
- [x] _* Return non-zero on failure of any task_
- [x] _* Ability to process all tasks even if one fails   (solved with `allowFailure`)_
- [ ] stash outstanding changes before running tasks, then restore the stash on completion

`*` Tasks with an asterisk are not fully satisfied by moon, because we can't get both of these features at the same time.  See below. 
 
That's really close!

As far as I can tell, moon immediately aborts all queued and unqueued tasks as soon as a failure is encountered (so long as the failed task does not have `allowFailure = true`).  For certain cases, like linting, failing fast is not desirable.  I would like an option to continue processing every task that does not have a failure in its upstream dependents. 

For example, consider a repo with two projects -- `foo` and `bar` -- with a set of global tasks like this:

```yaml
tasks:
  will_fail:
    command: 'false'
    options:
      cache: false

  will_succeed:
    command: 'true'
    options:
      cache: false

  lint:
    deps: [will_fail, will_succeed]
    command: "echo 'lint complete'"
```

When running `moon run :lint` I would like _all_ `will_fail` and `will_succeed` tasks to run.  This is safe because they are not dependent on each other. The `foo:lint` and `foo:bar` tasks would _not_ run their commands because they have failures in their deps (`foo:will_fail` and `bar:will_fail` respectively), so they are not safe to run.   

This behavior is particularly desirable for lint tasks, where a developer wants to perform or get feedback on a number of (usually unrelated) tasks as aggressively as possible.   I don't want all of my `check_yaml` tasks to abort just because one of my `check_python` tasks fail:  I want to know all the failures up front so that I can address all the issues together.

Instead, what currently happens is that tasks are aborted as soon as the first failure is detected, so I have to fix, rerun, and repeat to progressively expose issues, which is a waste of time. 

The closest that I can get to the desired behavior is to set `allowFailure = true` for my lint tasks. 

```yml
tasks:
  fails:
    command: 'false'
    options:
      cache: false
      allowFailure: true

  succeeds:
    command: 'true'
    options:
      cache: false
      allowFailure: true
```

There are two problems with this:

1.  I can no longer group these tasks together under another task, because tasks with `allowFailure = true` are not allowed to be depended upon.   This means that users need to provide the full list of lint tasks when calling `moon run`, or I need to create a wrapper script to do this for them.
2. If a task with `allowFailure = true` fails, `moon run` will still exit `0`.  Thus, when used as a precommit hook, it will not stop `git commit` from proceeding.  This makes it completely ineffective as a precommit solution.

**Option 1**:  global control over fail-fast behavior

One solution is a global option to disable the default fail-fast behavior.  `pre-commit` provides control over this behavior with a global option called [`fail_fast`](https://pre-commit.com/#top_level-fail_fast).  It defaults to false.  

Unlike the task-level `allowFailure` option, `moon run` should still return 1 if a task fails with `failFast = false`.  

This global option could be a CLI flag for `moon run` or an option in `.moon/workspace.yml`.

**Option 2**:  Add a task-level option to control failure propagation

A finer-grained solution could be to add an option to control the scope of failure propagation.

Possible semantics: 
- `abortTasksOnFailure` = "all", "downstream", or "none". Defaults to "all". 
- `failurePropagationScope` = "global", "project", "downstream", "none".  Defaults to "global".  

"none" is equivalent to  `allowFailure=true`.  Only "none" would cause moon run to exit 0 on task failure. 

Setting this in `taskOptions` within `.moon/tasks.yml` would have the effect of globally disabling the fail-fast behavior:

```yaml
taskOptions:
  abortTasksOnFailure:  downstream
```

If a task with `abortTasksOnFailure = all` fails it should not abort a task with `abortTasksOnFailure = downstream`. 

**Additional context**

chadrik:
The `--eager` flag makes moon run behave more like moon ci.

Fixes #1770

### DIFF
--- a/website/docs/commands/run.mdx
+++ b/website/docs/commands/run.mdx
@@ -42,6 +42,15 @@ CI, and what options are provided. The following scenarios are possible:
 
 :::
 
+:::info
+
+The default behavior for `moon run` is to "fail fast", meaning that any failed task will
+immediately abort execution of the entire action graph.  Pass `--no-bail` to execute as many
+tasks as safely possible (tasks with upstream failures will be skipped to avoid side effects).
+This is the default behavior for `moon ci`, and is also useful for pre-commit hooks.
+
+:::
+
 ### Arguments
 
 - `...<target>` - [Targets](../concepts/target) or project relative tasks to run.
@@ -59,6 +68,8 @@ CI, and what options are provided. The following scenarios are possible:
   [a query statement](../concepts/query-lang). <VersionLabel version="1.3.0" />
 - `--summary` - Display a summary and stats of the current run. <VersionLabel version="1.25.0" />
 - `-u`, `--updateCache` - Bypass cache and force update any existing items.
+- `-n`, `--no-bail` - When a task fails, continue executing other tasks instead of aborting immediately
+
 
 #### Affected
 
--- a/crates/app/src/commands/run.rs
+++ b/crates/app/src/commands/run.rs
@@ -53,6 +53,13 @@ pub struct RunArgs {
     )]
     pub update_cache: bool,
 
+    #[arg(
+        long,
+        short = 'n',
+        help = "When a task fails, continue executing other tasks instead of aborting immediately"
+    )]
+    pub no_bail: bool,
+
     // Debugging
     #[arg(
         value_enum,
@@ -201,7 +208,7 @@ pub async fn run_target(
     }
 
     // Process all tasks in the graph
-    run_action_pipeline(
+    let results = run_action_pipeline(
         session,
         ActionContext {
             passthrough_args: args.passthrough.to_owned(),
@@ -212,12 +219,24 @@ pub async fn run_target(
     )
     .await?;
 
+    if args.no_bail {
+        let failed = results.iter().any(|result| {
+            if result.has_failed() {
+                !result.allow_failure
+            } else {
+                false
+            }
+        });
+
+        if failed {
+            return Ok(Some(1));
+        }
+    }
+
     Ok(None)
 }
 
 #[instrument(skip_all)]
 pub async fn run(session: CliSession, args: RunArgs) -> AppResult {
-    run_target(&session, &args, &args.targets).await?;
-
-    Ok(None)
+    return run_target(&session, &args, &args.targets).await;
 }
--- a/crates/app/src/components.rs
+++ b/crates/app/src/components.rs
@@ -40,7 +40,7 @@ pub async fn run_action_pipeline(
             pipeline.summarize = true;
         }
         Commands::Run(cmd) => {
-            pipeline.bail = true;
+            pipeline.bail = !cmd.no_bail;
             pipeline.summarize = cmd.summary;
         }
         _ => {}
--- a/website/docs/config/workspace.mdx
+++ b/website/docs/config/workspace.mdx
@@ -731,7 +731,7 @@ name. [Learn more about Git hooks](https://git-scm.com/book/en/v2/Customizing-Gi
 vcs:
   hooks:
     pre-commit:
-      - 'moon run :lint :format --affected --status=staged'
+      - 'moon run :lint :format --affected --status=staged --no-bail'
       - 'another-command'
 ```
 


website/docs/commands/run.mdx
```diff
@@ -59,6 +66,8 @@ CI, and what options are provided. The following scenarios are possible:
   [a query statement](../concepts/query-lang). <VersionLabel version="1.3.0" />
 - `--summary` - Display a summary and stats of the current run. <VersionLabel version="1.25.0" />
 - `-u`, `--updateCache` - Bypass cache and force update any existing items.
+- `-e`, ``--eager` - When a task fails, continue executing other tasks instead of aborting immediately
```
milesj, line 69: double ticks here

chadrik:
I did a quick eval of a few other task runners to see if they have a similar option.  Here's the result:


| Tool | Flag |  Docs |
|--------|--------|--------|
| `just` | N/A |   (note: as far as I can tell `just` [runs tasks serially](https://github.com/casey/just?tab=readme-ov-file#running-tasks-in-parallel) |
| [`task`](https://github.com/go-task/task) | N/A |  |
| `turbo` | `--continue` |  Continue with task execution in the presence of an error (e.g. non-zero exit code from a task). When `--continue` is true, turbo will exit with the highest exit code value encountered during execution. Specifying [the --parallel flag](https://turbo.build/repo/docs/reference/run#--parallel) will automatically set --continue to true unless explicitly set to false. |
| [`pre-commit`](https://pre-commit.com) | `fail_fast` |   set to true to have pre-commit stop running hooks after the first failure. (note: pre-commit runs tasks serially) | 

It's unclear if `turbo run --continue` skips tasks with upstream failures like moon does, but it could be a good name for this feature.

website/docs/commands/run.mdx
chadrik: fixed

chadrik:
Note that all of the tools I've canvased so far default to eager/aggressive behavior, with a flag to manually enable the fail-fast mode.  As stated, turborepo defaults to `--continue` when in `--parallel` mode.

milesj:
> Note that all of the tools I've canvased so far default to eager/aggressive behavior, with a flag to manually enable the fail-fast mode. As stated, turborepo defaults to `--continue` when in `--parallel` mode.

Yeah I noticed this also. Would be a breaking change for moon, but something I can look into changing for 2.0.

chadrik:
I added the fix for exiting non-zero on failure.  

As for names, I prefer `--eager` because I feel that `--continue` could imply that even tasks with upstream failures will be run, but it is compelling that both Turborepo and Lage use `--continue`.  The decision is yours. 

> I'm leaning towards --continue myself (simple 1 word) but I also like the idea of supporting an enum.
> --on-failure bail (default), --on-failure continue, --on-failure 10 (bail after 10 errors)

I don't find the "bail after 10 errors" mode very useful, but I could see a trio of options like this being useful:

`--abort-on-failure`:  `immediate` (default),  `delayed` (process safe tasks), `disabled` (even tasks with upstream failures)

The final mode is not supported right now, but using an enum would keep the option open. Or just deal with it in 2.0.

The simplest solution is probably `-c` `--continue`.  Let me know and I'll make the adjustment.

milesj:
My only hesitation with `--eager` is that it reminds me of eager vs lazy loading, so I immediately think of "run all the things" instead of "keep running things".

Maybe `--no-bail` to keep it simple? I can revisit the enum pattern in v2.

chadrik:
Sounds good.  Should it be `--no-bail` or `--noBail`?

chadrik:
I pushed up the changes with `--no-bail`

milesj:
Awesome stuff, thanks! 

Aiming to release 1.31 on monday.

PR #1797, opened 2025-01-24
fix: Adapt remote git regex for templates

harlequin:
This change would allow to work with remote templates in a remote git repository which are hosted on different uri formats rather than github.

For example:

- `my-gitlab.cooperate-domain.com` - which was not working due to `-` in hostname
- `IDENT@my-gitlab.cooperate-domain.com` - which was not working due to `@` (for identity) and upper case letters 

Additional the automatic appending postfix `.git` to the git clone url is removed and is now in the user responsibility to add the correct url.

### DIFF
--- a/crates/codegen/src/codegen.rs
+++ b/crates/codegen/src/codegen.rs
@@ -210,8 +210,8 @@ impl<'app> CodeGenerator<'app> {
                     remote_url,
                     revision,
                 } => {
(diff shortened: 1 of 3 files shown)


milesj:
@harlequin Can you point it to the `develop-1.32` branch.

harlequin:
Point to develop 32

harlequin:
@milesj Test is fixed

PR #1834, opened 2025-02-17
fix: add cargo packages to package_names independently from aliases

alexander-heimbuch:


alexander-heimbuch:
Previously the list of implicit dependencies in a rust demo project was empty, I thought originally that this might be caused because of how the [workspace traversal works](https://github.com/moonrepo/moon/discussions/1802) by but while debugging this found this. Not sure if this intended, after moving it out of the condition it started to detect dependencies 🤷‍♂️

milesj:
@alexander-heimbuch Nice find! You'll also need to run `just bump patch` and commit the changes.

alexander-heimbuch:
@milesj done, I guess?

milesj:
@alexander-heimbuch Guess theres a formatting error

milesj:
Thanks!

PR #1836, opened 2025-02-19
docs(task-inheritance): merge example mixed command with args

sybernatus:
The doc on task inheritance is showing a merge between `command` and `args` between a global and a local task.

I think the correct example should be this one. What do you think?

### DIFF
--- a/website/docs/concepts/task-inheritance.mdx
+++ b/website/docs/concepts/task-inheritance.mdx
@@ -98,6 +98,7 @@ tasks:
   build:
     command:
       - 'webpack'
+    args:
       - '--mode'
       - 'production'
       - '--color'
@@ -126,6 +127,7 @@ tasks:
   build:
     command:
       - 'webpack'
+    args:
       - '--mode'
       - 'production'
       - '--color'


milesj:
The docs are actually correct, well both are. `command` supports an array of args, and it also gets merged with `args`. Internally, `command` because a string of just the 1st arg, but this is transparent to the user.

I'll leave the docs as-is, just to showcase other ways of doing things.

sybernatus:
My point is more about the illustration of `mergeArgs` options.

In the documentation, it's a little bit confusing because it seems that the option is appending `args` to `command`.
I don't think it is completely the case.

Let me illustrate a bit with `prepend` as it's more understandable like this:

---

From these templates: 

```yaml
# Global
tasks:
  build:
    command:
      - 'ls'
    args:
      - 'dir1'

# Local
tasks:
  build:
    args: 
      - 'dir2'
    options:
      mergeArgs: 'prepend'

```

I expect the result would be:
```yaml
# Merged result
tasks:
  build:
    command:
      - 'ls'
    args:
      - 'dir2'
      - 'dir1'
    options:
      mergeArgs: 'prepend'
```

Then `moon` run this as the following command `ls dir2 dir1`.

But the documentation seems to illustrate that `mergeArgs` is merging with command and would lead to this result:

```yaml
# Merged result
tasks:
  build:
    command:
      - 'dir2'
      - 'ls'
      - 'dir1'
    options:
      mergeArgs: 'prepend'
```

That I think is wrong.

Maybe I can add some example to the PR to make it more clear?

milesj:
I believe showing that `command` works with merging also is just as important as `args`, because more users use an array of values for `command` than separate `command` + `args`. If anything, I think the example should be updated to not use append and show a more complex example.

PR #1909, opened 2025-04-05
new: Add task `priority` option.

milesj:


rhuanbarreto:
This is so cool! Please let this be a number so we can have more control over levels here

PR #1933, opened 2025-04-24
[WIP] internal: Add ability to load non legacy toolchains

harlequin:
Add ability to load non legacy toolchains, like cpp toolchain for C development.

### DIFF
--- a/crates/app/src/systems/analyze.rs
+++ b/crates/app/src/systems/analyze.rs
@@ -142,6 +142,18 @@ pub async fn register_platforms(
     // Primarily for testing
     registry.reset();
 
+    for p in toolchain_config.get_enabled_platforms() {
+        registry.register(
+            p.get_toolchain_id(),
+            // TODO: Implement proper platform manager
+            Box::new(SystemPlatform::new(
+                workspace_root,
+                Arc::clone(proto_env),
+                Arc::clone(&console),
+            )),
+        );
+    }
+
     if let Some(bun_config) = &toolchain_config.bun {
         registry.register(
             PlatformType::Bun.get_toolchain_id(),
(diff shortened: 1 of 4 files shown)


harlequin:
@milesj: Can you have a look if this is the way how you would implement "dynamic" toolchain plugins.
The use case is that we need a toolchain for CPP development, and this is not in the "legacy" languages available.

Additional there is one todo left:

The CPP toolchain relies on conan, 
The wasm plugin is scanned for the proto plugin interfaces and moon interfaces, which is genius.
But the conan proto tool can only be done via the "shim" feature from proto.

Therefore, the task execution in moon will abort, because proto does not have the conan plugin location configured.

Do you have an idea, how to solve it best?

Solutions (what I have in mind)
1) proto will scan also for toolchain.yml instead only .prototools
2) Implement a "dynamic" way that the configured toolchain.yml plugin property will be "passed" into the proto call.

milesj:
All of this platform code will be going away soon, so your only option moving forward is to write a WASM plugin.

Examples here: https://github.com/moonrepo/plugins/tree/master/toolchains
And Rust which is currently in progress: https://github.com/moonrepo/plugins/pull/48

There's no docs an the APIs are still in flux, but they can be found here: https://github.com/moonrepo/moon/blob/master/crates/pdk-api/src/toolchain.rs

harlequin:
You are right, but the current Implementation does not Support Toolchain Plugins with other names, like a cpp Toolchain Name.

milesj:
Yes it does. You can give it any name in `toolchain.yml`, for example:

```yaml
# .moon/toolchain.yml
cpp:
  plugin: file://path/to/plugin.wasm
  someConfig: true
```

It gets resolved here: https://github.com/moonrepo/moon/blob/master/crates/config/src/toolchain_config.rs#L55

harlequin:
That is also correct, but than the action graph Bilder throws an error in the unwrap method. Line 826

With these changes it is working 

Give it a try

milesj:
The run task APIs for plugins aren't implemented yet, so that's expected.

Once this lands in v1.36 it should be good to go.

https://github.com/moonrepo/moon/pull/1934

PR #1957, opened 2025-05-08
adding skip_affected and skip dependents to node group

sbekel01:


milesj:
@sbekel01 What's this for?

PR #1965, opened 2025-05-12
feat(webhooks) Add acknowledge setting and global trace

### LINKED ISSUE #1961: [feature] Extend webhook functionalities
A great feature would be if the notification webhook will have 2 additional improvements:

1. Define that the webhook has to be acknowledged via 2xx return code
    
    Implementation would be as a configure option `notifier.acknowledge` with true or false inside the `workspace.yml` if webhooks are a fire and forget or wait for 2xx response.

2. Webhook `project.syncing` would include the `project` object from `moon.yml` configuration

    This would have the benefit if, for example, the data are stored in a Database, that Analysts can take the values and calculate runtimes over different projects.

3. Define global Trace ID via Environment variable.

    Currently it is implemented that every pipeline run will calculates it's own UUID for every pipeline run.
    In the scenario that we are splitting the jobs across different machines, `moon ci --jobs --jobsTotal`, it would be great that the user can calculate it's own "CI ID" before executing the moon commands. 

    With this the data can be analysed across machines.

@milesj: If you are agreeing that this is feature makes sense, I can provide the merge request.

harlequin:
Closes #1961

This enables enables the webhook to wait for acknowledge from remote receiver.

Additional via `MOON_TRACE_ID` a global trace id can be defined which will help to identify traces accross different machines when using **Parallelizing tasks**.


crates/notifier/src/webhooks.rs
```diff
@@ -78,31 +83,19 @@ impl WebhooksNotifier {
...
+            self.enabled = false;
+            self.verified = false;
+            warn!("Failed to send webhook event, subsequent webhook requests will be disabled");
+        } else {
             self.verified = true;
         }
-        // For every other event, we will make the request and ignore the result.
-        // We will also avoid awaiting the request to not slow down the overall runner.
-        else {
-            self.requests.push(tokio::spawn(async {
```
milesj, line 102: Should probably keep this if acknowledge is false, and inline it when true.

crates/action-pipeline/src/subscribers/webhooks_subscriber.rs
```diff
@@ -1,29 +1,44 @@
...
         }
     }
 }
 
 #[async_trait]
 impl Subscriber for WebhooksSubscriber {
     async fn on_emit<'data>(&mut self, event: &Event<'data>) -> miette::Result<()> {
-        self.notifier.notify(event.get_type(), event).await?;
+        if self.acknowledge {
+            let _ = self.notifier.wait_for_requests().await;
```
milesj, line 29: This shouldn't be necessary if the requests happen in the foreground.

crates/action-pipeline/src/subscribers/webhooks_subscriber.rs
```diff
@@ -1,29 +1,44 @@
...
+            if !self.notifier.verified {
+                self.receiver.terminate_running();
```
milesj, line 31: I think if you update `notify()` to return an error for non-200, it would just bubble up and shutdown the pipeline automatically.

milesj (commented):
Few questions

harlequin:
Hey Miles,
I dont think that I can handle it until monday. Sorry. 
We can plan it for next Iteration.

Sorry

milesj:
I merged the 1.36 branch and it auto-closed this.

Issue #1929, opened 2025-04-22
[feature] Provide environment variables to set defaults for certain moon run flags


chadrik:
**Is your feature request related to a problem? Please describe.**

As a developer, I'd like to create my own preference for certain `moon run` flags so that I don't have to provide them on the CLI each time. 

**Describe the solution you'd like**

Here are a handful of flags which users might want to enable in a semi-permanent way using environment variables:

```
--no-bail
--summary
--color
--dependents
--no-action
```

I propose we add the following environment variables:

```
MOON_NO_BAIL=1
MOON_SUMMARY=true
MOON_COLOR=True
MOON_DEPENDENTS=true
MOON_NO_ACTION=true
```

If you're open to this idea I can look into putting together a PR, unless you want to hammer it out real quick.  Just let me know.

**Describe alternatives you've considered**

users could create wrapper scripts around `moon run`, but that's a huge pain in the ass.

PR #1977, opened 2025-05-19
new: add environment variables for a few run arguments

chadrik:
Added environment variables to control `--summary`, `--no-actions`, `--no-bail`,  and `--affected`. 

Fixes https://github.com/moonrepo/moon/issues/1929

### DIFF
--- a/crates/app/src/commands/check.rs
+++ b/crates/app/src/commands/check.rs
@@ -21,6 +21,7 @@ pub struct CheckArgs {
     #[arg(
         long,
         short = 's',
+        env = "MOON_SUMMARY",
         help = "Include a summary of all actions that were processed in the pipeline"
     )]
     pub summary: bool,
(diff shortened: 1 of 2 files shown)


crates/app/src/app.rs
```diff
@@ -274,7 +274,7 @@ pub struct Cli {
     )]
     pub cache: CacheMode,
 
-    #[arg(long, global = true, help = "Force colored output")]
+    #[arg(long, global = true, env = "MOON_COLOR", help = "Force colored output")]
```
milesj, line 277: @chadrik This is actually wrong, since `MOON_COLOR` accepts more than just bools.

chadrik: Ok, I removed this.  Let me know if there's anything else you'd like me to do here.  I could update the help string to point users at the env var:  `Force colored output [env: MOON_COLOR=]`.

milesj:
Cool thanks! I'll merge through these errors, as theyre unrelated.

Issue #2008, opened 2025-06-12
Performance Issue in moon docker scaffold


aayushsinha44:
# Performance Issue: Duplicate Project Copying in `scaffold_sources_project` Due to Recursive Dependencies

## Problem Description
The `scaffold_sources_project` function in `crates/app/src/commands/docker/scaffold.rs` recursively copies project sources for dependencies, but doesn't handle duplicate dependencies efficiently. This leads to unnecessary file copying operations when projects have shared dependencies.

For example:
- Project A depends on B and C
- Project B also depends on C
- Current behavior: Project C's files are copied twice
- Expected behavior: Project C's files should be copied only once

## Impact
- Significantly increased execution time for `moon docker scaffold` command
  - **In our case we are seeing this command taking around 20 minutes. (Internally copying one library 105K times).** 
- Wastes disk space and I/O operations
- Doesn't scale well with large projects having many shared dependencies

## Current Implementation
The issue stems from the recursive implementation in `scaffold_sources_project`:

```rust
#[async_recursion]
async fn scaffold_sources_project(
    session: &MoonSession,
    project_graph: &ProjectGraph,
    docker_sources_root: &Path,
    project_id: &Id,
    manifest: &mut DockerManifest,
) -> AppResult {
    // ... 
    for dep_cfg in &project.dependencies {
        if !dep_cfg.is_root_scope() {
            scaffold_sources_project(
                session,
                project_graph,
                docker_sources_root,
                &dep_cfg.id,
                manifest,
            )
            .await?;
        }
    }
    // ...
}
```

Caching here will help. Don't process the same project twice.

## Additional Context
- File: [crates/app/src/commands/docker/scaffold.rs](https://github.com/moonrepo/moon/blob/master/crates/app/src/commands/docker/scaffold.rs)
- Function: `scaffold_sources_project` (around line 299)
- Related: The function is used by the `moon docker scaffold` command

aayushsinha44:
Created a fix here
https://github.com/moonrepo/moon/pull/2010

milesj:
Awesome thank you!

PR #2010, opened 2025-06-13
Fix: moon docker scaffold performance

aayushsinha44:
Fix for issue
https://github.com/moonrepo/moon/issues/2008

Currently `moon docker scaffold` command does a dependency graph traversal and copies the relevant files. But doesn't mark the node visited. Because of which same project gets processed twice i.e. copied twice.

We seen cases where this command takes more than 20 mins to run, where a single library gets copied over 105,000+ times. After this fix, the project now takes <30s for `moon docker scaffold` command to complete.

### DIFF
--- a/.yarn/versions/8992253b.yml
+++ b/.yarn/versions/8992253b.yml
@@ -1,0 +1,9 @@
+releases:
+  "@moonrepo/cli": patch
+  "@moonrepo/core-linux-arm64-gnu": patch
+  "@moonrepo/core-linux-arm64-musl": patch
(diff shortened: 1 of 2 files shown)


milesj:
@aayushsinha44 Looks good but you'll need to run `just bump patch` and commit the file.

aayushsinha44:
Thanks for the quick review @milesj. Pushed the `jump bump patch` file output`.

Also, may I know by when can we expect this to get released? This will help us improve our build time.

milesj:
Probably on monday, or tuesday the latest.

PR #2019, opened 2025-06-20
feat(webhooks) Add acknowledge setting and global trace (2nd attempt)

### LINKED ISSUE #1961: [feature] Extend webhook functionalities

harlequin:
feat(webhooks): Add acknowledge setting and global trace ID

This commit implements webhook acknowledgment and global trace ID features:
- Add webhookAcknowledge setting in workspace config
- Support MOON_TRACE_ID environment variable
- Update documentation

Closes #1961

### DIFF
--- a/crates/notifier/src/webhooks.rs
+++ b/crates/notifier/src/webhooks.rs
@@ -4,6 +4,7 @@ use moon_time::chrono::NaiveDateTime;
 use moon_time::now_timestamp;
 use serde::Serialize;
 use starbase_utils::json;
+use std::env;
 use tokio::task::JoinHandle;
 use tracing::{debug, trace, warn};
 use uuid::Uuid;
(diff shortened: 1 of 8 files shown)


harlequin:
@milesj: Next attempt to contribute for 1.38 ;)

Tell me if this is okay

crates/notifier/src/webhooks.rs
```diff
@@ -60,8 +72,10 @@ impl WebhooksNotifier {
             } else {
                 Uuid::new_v4().to_string()
             },
+            trace: env::var("MOON_TRACE_ID").unwrap_or("".into()),
```
milesj, line 75: I think `unwrap_or_default` would work here.

milesj:
@harlequin 1 comment and need to fix CI

harlequin:
@milesj should be okay now

milesj:
Awesme, looks good!

PR #2029, opened 2025-06-30
feat(config): support ron

rawkode:
This PR depends on a recent PR to schematic

https://github.com/moonrepo/schematic/pull/162


milesj:
@rawkode While I like this suggestion, I'm going to close this for now, since in v2 (by EOY), I will be reworking the config layer to support any format. Would rather wait till then for this.

PR #2065, opened 2025-07-24
docs: add mention of brew pkg

cedws:


milesj:
@cedws We don't maintain the brew package, so can you add a note that its unofficial.

PR #2087, opened 2025-08-25
fix: support git worktrees with bare repo

socksy:
The current logic is broken for the worktree layout I was trying to follow. I was vaguely trying to follow the following blog post's approach: https://morgan.cugerone.com/blog/how-to-use-git-worktree-and-in-a-clean-way/

It seemed a bit too brittle - always assuming there's going to be a .git dir, if I understand correctly. This change shortcircuits the worktree detection if it finds a .git file that has "gitdir:" in it, which to my belief only happens if we're in a worktree - so I think we're good to go on assuming we're in a worktree?

Basically, my belief is that the assumption that was underlying the existing logic is that we'll always be in a non bare repository, with some sub directory with worktrees in it. In that case, this logic seems sound. 

However, it doesn't work for the bare repository use case. And also, I believe worktrees can unfortunately be anywhere on the filesystem! Which means you can't really expect to be able to get to the real .git dir easily at all, but I don't think it's actually necessary for what it's used for in moon?

### DIFF
--- a/crates/vcs/src/gitx/git.rs
+++ b/crates/vcs/src/gitx/git.rs
@@ -78,8 +78,19 @@ impl Gitx {
                         "Found a .git file (submodule or worktree root)"
                     );
 
+                    // Check if this .git file represents a worktree
+                    let git_content = std::fs::read_to_string(&git_check)
+                        .map_err(|e| miette::miette!("Failed to read .git file: {}", e))?;
+
                     worktree_root = Some(current_dir.to_path_buf());
-                    // Don't break and continue searching for the actual root
+
+                    // If this is a worktree, stop traversal here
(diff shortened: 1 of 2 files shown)


crates/vcs/src/gitx/git.rs
```diff
@@ -78,8 +78,19 @@ impl Gitx {
                         "Found a .git file (submodule or worktree root)"
                     );
 
+                    // Check if this .git file represents a worktree
+                    let git_content = std::fs::read_to_string(&git_check)
```
milesj, line 82: Use `starbase_utils::fs::read_file` for fs stuff.

milesj:
I didn't close this on purpose, just merged the base and it auto-closed.

PR #2109, opened 2025-09-11
docs(vcs-hooks.mdx): use $@, which is compatible with unix-like shells and powershell

lucsorel:
the previously used `$1` syntax worked only on linux-like shells and broke the hook execution on windows computers with powershell.

`$@` works for both of them.

### DIFF
--- a/website/docs/guides/vcs-hooks.mdx
+++ b/website/docs/guides/vcs-hooks.mdx
@@ -33,7 +33,7 @@ vcs:
       - 'pre-commit run'
       - 'moon run :lint --affected'
     commit-msg:
-      - 'pre-commit run --hook-stage commit-msg --commit-msg-filename $1'
+      - 'pre-commit run --hook-stage commit-msg --commit-msg-filename $@'
 ```
 
 :::info


milesj:
@lucsorel Isn't `$@` the list of all args? And not just the first?

lucsorel:
> Isn't `$@` the list of all args? And not just the first?

Yes it is. But the `commit-msg` hook seems to be called by git with only one parameter, the filename containing the commit message (often `.git/COMMIT_EDITMSG`). The command works because the 1st value (at index 0) - which is the full command - is not printed when echoed.

I modified the `.moon/hooks/commit-msg.sh` file:

```sh
#!/usr/bin/env bash
set -eo pipefail

# Automatically generated by moon. DO NOT MODIFY!
# https://moonrepo.dev/docs/guides/vcs-hooks

echo "args: $@"
echo "args 0: ${@:0}"
echo "args 1: ${@:1}"
uv run pre-commit run --hook-stage commit-msg --commit-msg-filename $@
```

And it produced the following output (which surprises me):

```sh
args: .git/COMMIT_EDITMSG
args 0: ./.moon/hooks/commit-msg.sh .git/COMMIT_EDITMSG
args 1: .git/COMMIT_EDITMSG
# and pre-commit worked fine
```

I will check tomorrow with a colleague using a windows computer whether powershell supports the `${@:1}` syntax (but I doubt it). If it works with powershell, I will modify the documented command, if that is also what you prefer.

lucsorel:
hi @milesj :slightly_smiling_face: 

> I will check tomorrow with a colleague using a windows computer whether powershell supports the `${@:1}` syntax (but I doubt it). If it works with powershell, I will modify the documented command, if that is also what you prefer.

We checked: powershell does not support the `${@:1}` syntax. So I suggest that we stick with the given commit, which works both in unix-like shells (zsh on Ubuntu) and powershell (5.1.22621 revision 2428, with Windows 11 version 24H2).

**Further thinking about moon sync hooks**

When using `moon sync hooks`, soon I find it hard to define hook commands (in the `vcs.hooks` section of `.moon/workspace.yml`) that are compatible for different shells (linux ones and powershell). I was wondering if it were possible to commands specific to the shell environment? Something like:

```yaml
vcs:
  manager: 'git'
  defaultBranch: 'main'
  hooks:
    pre-commit:
      shell:
        - 'echo "${@:1}"'
      powershell:
        - 'echo "$args[1]"'
```

Although I don't think it is such a good idea when using the `pre-commit` tool, since it already handles this multi-shell pain with the [`pre-commit install` command](https://pre-commit.com/#3-install-the-git-hook-scripts), which also generates the `.git/hooks/...` scripts compatible with the local shell.

An alternative to the automatic activation of hooks with `moon sync hooks` would be the possibility to run a user-defined command (`pre-commit install`, in this case) when `moon` synchronizes the workspace. Is this an available feature of moon?

It could be a root task of the workspace, defined as a dependency of all other tasks, but that would be painful to write for the users.

milesj:
I can probably just have it set env vars that can be accessed, like `$ARG1`, etc.

lucsorel:
> I can probably just have it set env vars that can be accessed, like `$ARG1`, etc.

yes. Before that, I believe the documentation should be fixed so that it works for both unix-based shells and powershell. What do you think?

lucsorel:
> I can probably just have it set env vars that can be accessed, like `$ARG1`, etc.

I see that you are addressing this issue in #2116 :smiley: , thank you :pray:

milesj:
@lucsorel yeah added, give 1.40.3 a try

lucsorel:
Version https://github.com/moonrepo/moon/releases/tag/v1.40.3 fixed the issue, thank you :pray:

I thus close this PR. Thank you for your reactivity

PR #2117, opened 2025-09-18
feat: Detect subcommand and apply shell on them

royreznik:
I have encounter an annoying behaviour when trying to use moon

When i added the following command to my target:
```
command: docker run --rm -e HOME="/tmp" -u "$(id -u):$(id -g)"  ...
````

It didn't automaticly changed it into shell, like moon does when it sees a '*', as for `if task.args.iter().any(|arg| is_glob_like(arg))` 

So i thought it will be a good idea to just set it into shell in those cases

### DIFF
--- a/crates/env-var/src/env_substitutor.rs
+++ b/crates/env-var/src/env_substitutor.rs
@@ -25,9 +25,53 @@ pub static ENV_VAR_SUBSTITUTE_BRACKETS: LazyLock<Regex> = LazyLock::new(|| {
     .unwrap()
 });
 
+// $(command)
+pub static SUBCOMMAND_SUBSTITUTE: LazyLock<Regex> = LazyLock::new(|| {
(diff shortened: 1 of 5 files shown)


milesj:
@royreznik `script` instead of `command` was more intended for this kind of use case.

royreznik:
> @royreznik `script` instead of `command` was more intended for this kind of use case.

Thanks for the reply
This just to make it consistent, as we already have it for globs, lets do it for subcommands

milesj:
The problem with `command` is that it is constantly string split and re-joined (`script` is not), so there's no guarantee that the command will actually be pieced back together correctly again. This will probably not be allowed in v2.

So I'm not sure I want to fully support shell syntax in `command`.

milesj:
I wrote a command line parser that better handles this going forward in v2.

https://crates.io/crates/starbase_args

PR #2157, opened 2025-10-02
fix: exclude RunInCI 'only' tasks from `moon check`

invokermain:


milesj:
While I appreciate the effort, `runInCI` is a `moon ci` only option right now, so this diverges from that behavior.

This will be changing in v2 where run/check/ci commands all share the same logic.

invokermain:
No problem, I can go back to the CI env var workaround.

So am I understanding it correctly - the behaviour of `runInCI: only` is correct here:

```
        match (self.ci, &task.options.run_in_ci) {
            (true, TaskOptionRunInCI::Always) => {
                return Ok(Some(AffectedBy::AlwaysAffected));
            }
            (true, TaskOptionRunInCI::Enabled(false))
            | (true, TaskOptionRunInCI::Skip)
            | (false, TaskOptionRunInCI::Only) => {
                return Ok(None);
            }
            _ => {}
        };
```

its just that `moon check` uses a different code path (i.e. does not use the `AffectedTracker`) and you plan to align this in 2.0 at which point it will work as intended?

milesj:
Yah basically.

PR #2229, opened 2025-11-24
Add moonx binary entry to package.json

jamestalmage:
Link moonx for folks who install via a node package manager.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -18,7 +18,8 @@
   "author": "Miles Johnson",
   "license": "MIT",
   "bin": {
-    "moon": "moon.js"
+    "moon": "moon.js",
+    "moonx": "moonx.js"
   },
   "repository": {
     "type": "git",


milesj:
@jamestalmage I actually had this before but it breaks the `moon` binary, since npm is dumb and doesn't know what the "default" bin is anymore.

npm's logic is stupid, as the default bin name === the package name, and since the package name isn't "moon" this won't work.

jamestalmage:
I'm doing this (installing multiple binaries) in internal command line packages, and it works fine in `npm 11.2`

jamestalmage:
Confirmed that this works for moon.

```
npm install @jamestalmage/moonrepo-cli
```

Running moonx gives expected results

image

milesj:
I wonder if they fixed this in latest npm. Will need to also verify older versions of npm, and then pnpm/yarn.

jamestalmage:
I'm pretty sure this has worked in npm for a while. Like a year or more... Any chance your test-suite also uses moon (i.e. dogfooding, etc), and the actual issue related to https://github.com/moonrepo/moon/issues/2270?

jamestalmage:
yahoo

PR #2249, opened 2025-12-05
correct past participle in intro.mdx

brandonzylstra:
"...can still be run" is the standard English way of saying this, because "run" is a past participle and "ran" is the simple past tense. Even though it is becoming more common to replace the past participle with the simple past tense, it is still generally considered an error by a majority of native speakers.

Here's some more information, although these may focus more on examples involving perfect tenses rather than passive voice (which is in use above):
- [An academic study of this phenomenon](https://www.academia.edu/110304820/Where_Have_All_the_Participles_Went_Using_Twitter_Data_to_Teach_About_Language)
- [Observations by some language bloggers](https://grammarphobia.com/blog/2011/12/participles.html)
- [An AI summary](https://duckduckgo.com/?q=where+have+all+the+participles+went%3F&ia=web&assist=true)


milesj (approved):

milesj:
Thanks

Issue #2253, opened 2025-12-09
[feature] add `--remote` to `moon query projects` command


miguelvr:
**Is your feature request related to a problem? Please describe.**

When working with moon in CI/CD pipelines, I need to query project information from the remote diffs rather than relying solely on local state. Currently, `moon query projects` only operates on local data, which means this is not possible without a seriously hacky script.

**Describe the solution you'd like**

Add a `--remote` flag to the `moon query projects` command, consistent with how other moon commands support remote diffs. This would allow querying affected files from already committed changes.

Example usage:
```bash
moon query projects --remote
moon query projects --remote --json
```

**Describe alternatives you've considered**

- Checking out the previous base commit before running queries, but this adds unnecessary overhead and complexity that I would like to avoid as a user

**Additional context**

This would bring `moon query projects` in line with other commands that already support the `--remote` flag, providing a more consistent CLI experience.

miguelvr:
closing this since it's not going to be accepted as a change to v1

for posteriority, there is a workaround by setting the `MOON_BASE` environment variable:

```bash
# Determine the base commit for comparison
# - For PRs: use the PR base branch (GITHUB_BASE_REF)
# - For main branch: use HEAD~1
# - For local development: use origin/main
if [ -n "${GITHUB_BASE_REF:-}" ]; then
  # Pull request - compare against the PR base branch
  export MOON_BASE="origin/${GITHUB_BASE_REF}"
elif [ "${GITHUB_REF:-}" = "refs/heads/main" ]; then
  # Main branch - compare against previous commit
  export MOON_BASE="HEAD~1"
else
  # Local development or other branches - compare against origin/main
  export MOON_BASE="origin/main"
fi
moon query projects --affected --downstream=deep
```

PR #2254, opened 2025-12-10
feat: add support for `--remote` on moon query (task|projects)

miguelvr:
Closes https://github.com/moonrepo/moon/issues/2253


milesj:
v2 is landing soon and this will be changing, so not landing v1 changes right now. https://github.com/moonrepo/moon/pull/2181

PR #2283, opened 2026-01-04
fix: correct issue where the global version is not delegating to the local version

Playjasb2:
There was a strange behaviour where if you have a global version and a packaged version of moon installed, moon commands you execute will not delegate over to the local version.

The root cause seems to be in the `has_locally_installed` function where it tries to look for the `moon` binary on Linux/macOS or `moon.exe` on Windows as part of the packaged version lookup.

However, in a pnpm workspace setup I am using, I do not see the binary:

Screenshot 2026-01-03 at 6 57 09 PM

It is to my knowledge that the `moon` binary would reside in a platform-specific package like `@moonrepo/core-macos-arm64`, but as you can see in my `node_modules` directory, it may not appear at the same level as my other dependencies; it resides in `node_modules/.pnpm/`.

The fix for this is to directly check for the binary from the platform's package, and resolve any symlinks when needed.

### DIFF
--- a/crates/cli/src/lookup.rs
+++ b/crates/cli/src/lookup.rs
@@ -94,18 +94,16 @@ pub fn has_locally_installed(home_dir: &Path, current_dir: &Path) -> Option<Path
 
                 // 2. Real Path (pnpm, Bun, Yarn Berry)
                 // If it's a symlink, resolve it to find the real location in the store or cache.
-                if let Ok(real_path) = fs::canonicalize(&cli_dir) {
-                    if real_path != cli_dir {
-                        search_paths.push(real_path.clone()); // Inside the package itself
-                        if let Some(parent) = real_path.parent() {
-                            search_paths.push(parent.to_path_buf()); // Sibling in store
-                        }
-                        if let Some(grandparent) = real_path.parent().and_then(|p| p.parent()) {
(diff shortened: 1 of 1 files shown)


milesj:
Ah yeah this is an oversight. We previously had a postinstall script that would copy the moon(.exe) into the package, so this check used to work just fine. But we removed the postinstall script because of Bun.

milesj:
@Playjasb2 Looks good, just need to fix lints.

Playjasb2:
I fixed the linting issue. It should be good to go.

Playjasb2:
The aarch64 Linux builds are failing because it ran out of memory. Fat LTO seems to take up too much memory on those platform.

One suggestion is that you could make PR builds not use LTO to speed things up, and then have LTO be applied towards the actual releases.

But as for the actual release builds, you could opt for thin LTO, which many in the community do, as it's faster to compile with and uses less memory and you can achieve performance similar to fat LTO.

However it's understandable that you would want the absolute fastest and space efficient builds, which fat LTO can do, but sadly for the aarch64 Linux platforms, you would need more than 16GB of RAM, which isn't free. I tried using lld, but even with that, I still get the same error. I'm not sure if there are any other techniques to try to avoid running out of memory in this case.

milesj:
I tweaked the settings, try rebasing.

PR #2298, opened 2026-01-14
Fix Boolean logic for `OR` clause applied to empty list

agdphd:
The current logic for inherited clause logic in Moon has surprising behavior in a corner case: given a clause `OR(...conditions)`, it says that the clause does *not* match the empty list `[]`!

This came up for me in the context of task inheritance, where I wanted to configure a linter task which would be disabled with a tag. It seems natural to configure the task with: `inheritedBy: {tags: { not: 'nolint' } }` and then set `tags: ['nolint']` on the relevant projects, but this only works *if the project has another tag set*.

This turns out to be a pretty simple fix, so here it is!


milesj:
@agdphd Can you rebase and see if that fixes the flaky tests.

milesj:
Hrmm this errors are unrelated but they are also on master. Will merge through and fix them there.

PR #2300, opened 2026-01-15
Pass through error messages in OneOfMany deserialization

Adjective-Object:
I was trying to figure out if I could use globs to filter task inheritance within a repo. This is not allowed, but the error messages I was being presented with were confusing:

This change fixes that by taking an approach similar to #2292. In order to maintain the flexibility to support `OneOf<Vec<T>>` and avoid future footguns, this change replicates the backoff-approach that `serde-derive` takes with its derive macro for untagged enums. However, because access to serde's  `ContentRefDeserializer` is no longer publicly accessible, this change adds `serde-content` to achieve the same thing.

### Previously
```
$ moon run '*:mytask'
Error: config::parse::failed

  × Failed to parse .moon/tasks/migration.yml.
  ╰─▶   × inheritedBy.files: expected a single value, or a list of values
      
  help: https://moonrepo.dev/docs/config/tasks

```

### Now:
```
$ moon run '*:vitest-migrate'
Error: config::parse::failed

  × Failed to parse .moon/tasks/migration.yml.
  ╰─▶   × inheritedBy.files[0]: globs are not supported, expected a literal file path
      
  help: https://moonrepo.dev/docs/config/tasks

```

## Context
I'm working in a monorepo and running a large migration script across all packages that contain a specific kind of integration test file. I'm using moonrepo to coordinate the migration.

The migration script is slow, we don't control it, and coverage using this specific kind of test is sparse.

I was checking if it would be possible to filter task inheritence for the "migrate" task using globs in order to shrink the build graph, and ran into this confusing error message and looked deeper.

### DIFF
--- a/crates/config/src/shapes/poly.rs
+++ b/crates/config/src/shapes/poly.rs
@@ -185,7 +185,6 @@ mod test {
         Deserialize,
         de::{Deserializer, Error},
     };
-    use serde_json;
     use std::fmt::Debug;
 
     pub fn custom_err<'de, D, T>(_deserializer: D) -> Result<T, D::Error>


milesj:
I wonder if I should update https://github.com/moonrepo/schematic for untagged enums to always bubble up the error. That way we can squash this across the board.

Adjective-Object:
Tt could feasibly be addressed in config_enum, but the specifics of how error messages get propagated would have to be made more generic. Something like:

```
failed to parse as any variant of the enum:
- One<T>: (error content)
- Many<T>: (error content)
```

Seems like the easiest solution that addresses it within `schematic`. I'll give it a shot sometime

Adjective-Object:
Superseded by https://github.com/moonrepo/schematic/pull/183

PR #2314, opened 2026-01-20
docs: Add fish shell install command

yuhua99:
Adds a `psub` example for Fish users.
The standard `<(...)` syntax is not supported in Fish, and simple piping (`| bash`) breaks the interactive prompts.

Fixes https://github.com/moonrepo/proto/issues/905


milesj:
Thanks!

PR #2316, opened 2026-01-20
feat: add --hide-cached

### LINKED ISSUE #1930: [feature] Provide an option for moon run to hide output from cached tasks

As a developer, I would like to prevent `moon run` from printing output from cached tasks because A) it can be misleading by giving the impression that the task is running, and B) some tasks have a lot of output and it makes it harder to see the output of uncached tasks, which are more important. 

A new flag for `moon run` to control this behavior.

I can't think of any alternatives. 

**Additional context**

I experimented with `--quiet` and noticed that it overrides `--summary`.   I think it would be useful to combine these: i.e. use `--quiet` to remove some of the extraneous ▪▪▪▪ markers, but still preserve the final summary.  I can make a separate ticket for this.

Adjective-Object:
Adds --hide-cached to the `exec` family of commands. This is used to configure the MoonReporter's behaviour for a given command run.

At present the only way to configure task reporting is with the individual task's reporting settings. This change adds a new mechanism to configure the reporter behaviour from command line flags.

A "clean" separation of concerns might complicate this by making a separate MoonTaskReporter that is aware of tasks, and initialize it in each of the exec-family commands. However, because the MoonReporter is aware of tasks and configured before subcommand execution, we need to read the CLI commands outside of the individual subcommands in order to initialize it before descending into subcommand execution.

I'm not 100% happy with this design, but I tried plumbing overrides through TaskReportItem, and I didn't like that design because it conflated transient, global CLI config with persistent config from individual task configs. I think this is a reasonable compromise that doesn't involve rewriting the MoonReporter to be more "pure" and harder to maintain.

Closes #1930

### DIFF
--- a/crates/app-macros/src/lib.rs
+++ b/crates/app-macros/src/lib.rs
@@ -48,6 +48,17 @@ pub fn with_shared_exec_args(attr: TokenStream, item: TokenStream) -> TokenStrea
             )]
             pub summary: Option<Option<crate::app_options::SummaryOption>>
         },
+        quote! {
+            #[arg(
+                long,
+                short = 'x',
+                global = true,
+                env = "MOON_HIDE_CACHED",
+                help = "Hide all stdout/stderr from succesful cached tasks",
+                help_heading = super::HEADING_WORKFLOW,
+            )]
+            pub hide_cached: bool
+        },
         // GRAPH
         quote! {
             #[arg(
--- a/crates/app/src/app.rs
+++ b/crates/app/src/app.rs
@@ -351,6 +351,28 @@ pub struct Cli {
 }
 
 impl Cli {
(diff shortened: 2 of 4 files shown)


Adjective-Object:
In my perfect world there would be a way to configure this at the repo level as a default setting as well, but I couldn't figure out if that matches with any of the existing config systems.

Is there a decent place to discuss these things out of band? I keep seeing people mention a Discord in other issues/PRs but can't find it online anywhere 😅

milesj:
@Adjective-Object Sorry for the late reply, wanted to get v2 out.

I've though about this before as well, and like you've seen, it's kind of awkward to get the settings into the right code paths.

This may make more sense as a config setting so that users don't always have to pass the arg. Maybe `pipeline.hideCachedTaskOutput` or something?

Issue #1828, opened 2025-02-11
[feature] Ability to run tasks for the current project without having to specify the project name


rdvoskin-oneapp:
**Is your feature request related to a problem? Please describe.**

Currently when running any task on a specific project, the name of the project must be specified.  For example for `app` to run `build` we do
```
  moon app:build
``` 

**Describe the solution you'd like**

It would nice to be able to omit the project name if the current directory has a moon.yml file. This is supported in [rush](https://rushjs.io/pages/commands/rushx/#rush-vs-rushx) via `rushx`. For the moon case it would nice to have a `moonx` command. For example
```
moonx :build
```

**Describe alternatives you've considered**

I've considered writing my own bash alias that parses the `moon.yml` file in the working directory and executes 
```
moon $current_project_name:$task
```

**Additional context**

milesj:
@rdvoskin-oneapp This is already possible but requires `run`: https://moonrepo.dev/docs/commands/run

So `moon run build` would run the `build` task in the closest project. We can't use the shorthand (no `run`) here, because the `:` is used as a marker when rewriting the args to make shorthand possible.

However, we could probably make `moon ~:build` handle this scenario.

rdvoskin-oneapp:
I think something like `~:<task>` is a great alternative to creating an alias like `moonx`! Good to know that `moon run` finds the closest project. One advantage of an alias like `moonx` is it could solve the `run` issue (re-writing args) and allow something that's even shorter like `moonx build` (no `~` and no `:`).

milesj:
Yeah `moonx` is a cool idea too. I'll look into it.

milesj:
Landed in v1.33.

PR #2370, opened 2026-02-20
docs: Add "Closest project" scope in run and exec command pages

agarleanu:
I have never used the closest scope before, but intuitively I assumed moon must have a feature for running tasks in the current working directory.

I looked through issues and found https://github.com/moonrepo/moon/issues/1828, which thankfully also had a linked [PR](https://github.com/moonrepo/moon/pull/1853/).

This very useful command is hard to find by just reading the documentation.

This may not be the best place to mention this scope, but since those pages contain quick summaries of those commands (run & exec), I assumed it would be useful to have the "closest" scope mentioned there.

Let me know if you'd also like me to re-write some other section of the documentation to make this clearer to users.

### DIFF
--- a/website/docs/commands/exec.mdx
+++ b/website/docs/commands/exec.mdx
@@ -27,8 +27,8 @@ $ moon exec :test
 $ moonx :test
 
 # Run `test` in the closest project, relative to the current working directory
-$ moon exec ~:test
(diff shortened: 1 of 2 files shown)


milesj:
Thanks!

PR #2393, opened 2026-03-03
Build for Intel Macs

agdphd:
Some folks at my company still use Intel MacBooks, and they'd like to be able to use Moon 😄


milesj:
We dropped Intel support because GH is phasing it out, and Apple no longer supports it.

agdphd:
> We dropped Intel support because GH is phasing it out, and Apple no longer supports it.

... ah, that makes sense 😅

PR #2407, opened 2026-03-09
fix: Emit valid JSON Schema for MCP generate tool variables parameter

### LINKED ISSUE #2396: [bug] MCP server emits invalid JSON Schema "type":"unknown" for generate tool
**Describe the bug**

The `moon mcp` server returns `"type": "unknown"` in the JSON Schema for the `generate` tool's `variables` parameter. `"unknown"` is not a valid JSON Schema type per draft 2020-12, which only allows `string`, `number`, `integer`, `boolean`, `array`, `object`, and `null`.

This causes clients that validate tool schemas (e.g. Claude Code / Anthropic API) to reject the entire toolset with:

```
API Error: 400 tools.4.custom.input_schema: JSON schema is invalid.
It must match JSON Schema draft 2020-12.
```

**Steps to reproduce**

1. Run `moon mcp`
2. Send an MCP `initialize` request followed by `tools/list`
3. Inspect the `generate` tool's `inputSchema` — the `variables` field has `"type": "unknown"`

**Expected behavior**

The `variables` field should use a valid JSON Schema type, e.g. `"type": "object"`.

**Environment**

- moon 2.0.4
- macOS Darwin 25.0.0
- Node 22.x

**Additional context**

Workaround: use this wrapper script instead of `moon mcp` directly. It proxies stdin/stdout and patches the invalid schema type on the fly. Requires Node.js.

```bash
#!/usr/bin/env bash
# Wrapper to launch moon mcp with schema fixes applied.
# Moon 2.0.4 emits "type":"unknown" in the generate tool's input_schema,
# which is invalid JSON Schema and causes Claude's API to reject all tools.
# This proxy patches the response on the fly until the bug is fixed upstream.
set -euo pipefail

export MOON_WORKSPACE_ROOT="${MOON_WORKSPACE_ROOT:-$(git rev-parse --show-toplevel)}"

exec node -e '
const { spawn } = require("child_process");
const proc = spawn("moon", ["mcp"], { stdio: ["pipe", "pipe", "inherit"] });

process.stdin.pipe(proc.stdin);

let buf = "";
proc.stdout.on("data", (chunk) => {
  buf += chunk.toString();
  let idx;
  while ((idx = buf.indexOf("\n")) !== -1) {
    const line = buf.slice(0, idx);
    buf = buf.slice(idx + 1);
    if (line.trim()) {
      const fixed = line.replace(/"type"\s*:\s*"unknown"/g, "\"type\":\"object\"");
      process.stdout.write(fixed + "\n");
    }
  }
});

proc.stdout.on("end", () => {
  if (buf.trim()) {
    const fixed = buf.replace(/"type"\s*:\s*"unknown"/g, "\"type\":\"object\"");
    process.stdout.write(fixed);
  }
});

proc.on("exit", (code) => process.exit(code ?? 1));
process.on("SIGTERM", () => proc.kill("SIGTERM"));
process.on("SIGINT", () => proc.kill("SIGINT"));
'
```

Point your MCP client configuration at this script instead of running `moon mcp` directly.

gabrielbryk:
## Summary

- The `rust-mcp-macros` `JsonSchema` derive emits `"type": "unknown"` for generic types like `FxHashMap<String, Value>`, which is invalid per all JSON Schema drafts. This causes the Anthropic API (and other strict validators) to reject all tool registrations with a 400 error.
- Introduces a `Variables` newtype wrapper with a hand-written `json_schema()` method returning `{"type": "object"}`, preserving serde compatibility via `#[serde(transparent)]` and `Deref`.
- Adds integration tests validating the generated schemas against JSON Schema Drafts 4, 6, 7, 2019-09, and 2020-12.

Closes #2396

## Test plan

- [x] `Variables::json_schema()` produces a valid schema across all 5 JSON Schema drafts
- [x] `Variables` serde roundtrip (serialize → deserialize) preserves data
- [x] `Generate::json_schema()` is valid across all 5 drafts
- [x] `Generate::tool().input_schema` is valid across all 5 drafts
- [x] Sample input validates against the tool input schema
- [x] Missing required fields are rejected by the tool input schema
- [x] Verified regression: reverting fix to `"unknown"` causes 5/6 tests to fail

### DIFF
--- a/crates/mcp/src/tools/codegen_tools.rs
+++ b/crates/mcp/src/tools/codegen_tools.rs
@@ -10,6 +10,33 @@ use rust_mcp_sdk::{
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::ops::Deref;
+
+/// Newtype wrapper around `FxHashMap<String, Value>` that provides a valid
+/// JSON Schema (`"type": "object"`) instead of the `"type": "unknown"` that
+/// the `rust-mcp-macros` `JsonSchema` derive emits for unrecognized generic types.
+#[derive(Debug, Default, Deserialize, Serialize)]
+#[serde(transparent)]
+pub struct Variables(pub FxHashMap<String, serde_json::Value>);
+
+impl Variables {
+    pub fn json_schema() -> serde_json::Map<String, serde_json::Value> {
+        let mut map = serde_json::Map::new();
+        map.insert(
+            "type".to_string(),
+            serde_json::Value::String("object".to_string()),
+        );
+        map
+    }
+}
+
+impl Deref for Variables {
+    type Target = FxHashMap<String, serde_json::Value>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.0
+    }
+}
 
 #[mcp_tool(
     name = "generate",
@@ -28,7 +55,7 @@ pub struct Generate {
     pub force: Option<bool>,
 
     #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub variables: Option<FxHashMap<String, serde_json::Value>>,
+    pub variables: Option<Variables>,
 }
 
 impl Generate {
@@ -54,7 +81,7 @@ impl Generate {
         {
             args.push("--".into());
 
-            for (key, value) in vars {
+            for (key, value) in vars.iter() {
                 let opt = format!("--{key}");
 
                 match value {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -26,6 +26,7 @@ dependencies = [
  "cfg-if",
  "getrandom 0.3.4",
  "once_cell",
+ "serde",
  "version_check",
  "zerocopy",
 ]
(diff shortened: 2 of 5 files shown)


crates/mcp/src/tools/codegen_tools.rs
```diff
@@ -10,6 +10,33 @@ use rust_mcp_sdk::{
 use rustc_hash::FxHashMap;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::ops::Deref;
+
+/// Newtype wrapper around `FxHashMap<String, Value>` that provides a valid
```
milesj: @gabrielbryk Does it work correctly if we change `FxHashMap` to the plain `HashMap`?

gabrielbryk: No, unfortunately `HashMap` would have the same issue. The `rust-mcp-macros` [`type_to_json_schema()`](https://github.com/rust-mcp-stack/rust-mcp-sdk/blob/main/crates/rust-mcp-macros/src/utils.rs) function only has special handling for `Option`, `Vec`, `String`, `bool`, and numeric primitives. Any other generic type (including `HashMap<K, V>`) fails the `might_be_struct()` check (because it has generic arguments) and falls through to the `"unknown"` fallback.

Their own test suite confirms this:
```rust
let ty: syn::Type = parse_quote!(std::collections::HashMap<String, i32>);
assert!(!might_be_struct(&ty)); // segments.len() > 1
```

So the newtype wrapper is needed regardless of which map type is used underneath.

PR #2483, opened 2026-04-15
fix: reindex filtered workspace graphs

gabrielbryk:
## Summary
After upgrading to `2.2.0` in another project that uses moon, CI started timing out during graph build. I traced the regression back to filtered workspace graph reindexing in moon and fixed it here.

This PR:
- rebuilds project/task side maps from post-filter node indices so filtered workspace graphs keep correct direct dependencies
- adds regression coverage in the workspace builder for filtering `Loading` nodes
- adds higher-level partial-load regression tests for public project/task dependency queries

## Root cause
`#2469` reduced graph memory usage by changing project/task/workspace graphs to store `NodeIndex` weights plus side maps instead of full node payloads.

During workspace graph finalization, `filter_map` drops `Loading` nodes and may reindex the DAG. The side maps were still keyed by the pre-filter indices, so `ProjectGraph` and `TaskGraph` could resolve the wrong node for a given index. That broke direct dependency queries like `dependencies_of()`, which then corrupted action graph sync requirements.

## What changed
- rebuild `ProjectGraph.indexes` and `ProjectGraph.nodes` from the filtered graph's new indices
- rebuild `TaskGraph.indexes` and `TaskGraph.nodes` the same way
- apply the same fix in both the sync workspace builder and the shared async finalize paths

## Verification
- `cargo test -p moon_workspace reindexes_ -- --nocapture`
- `cargo test -p moon_project_graph --test project_graph_test isolation -- --nocapture`

### DIFF
--- a/crates/project-graph/tests/project_graph_test.rs
+++ b/crates/project-graph/tests/project_graph_test.rs
@@ -1001,11 +1001,15 @@ mod project_graph {
                 let mock = create_workspace_mocker(sandbox.path());
 
                 let graph = mock.mock_workspace_graph_for(&["some-depends-on"]).await;
+                let project = graph.get_project("some-depends-on").unwrap();
+                let mut direct_deps = map_ids(graph.projects.dependencies_of(&project));
+                direct_deps.sort();
 
                 assert_eq!(
                     map_ids(graph.projects.get_node_keys()),
                     ["some-depends-on", "a", "c"]
                 );
+                assert_eq!(direct_deps, ["a", "c"]);
             }
 
             #[tokio::test(flavor = "multi_thread")]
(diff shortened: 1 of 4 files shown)


milesj:
@gabrielbryk The filtering out of `Loading` was also happening in previous versions, so that's nothing new.

What situation were you hitting this issue because finalize _should_ have all the nodes in the graph by that point.

gabrielbryk:
Right, the filtering itself isn't new.

The issue is that after `#2469`, filtering stopped being harmless: the graph now stores `NodeIndex` weights plus side maps, so if `filter_map` reindexes the DAG and those side maps still point at the old indices, dependency lookups start resolving to the wrong projects/tasks.

On the second point, I agree that the normal full-graph path should ideally have everything loaded by `finalize()`. The bug here is in the filtered finalization paths themselves, where `Loading` nodes can still exist, for example in focused/partial graph builds. In those cases, the old indices were being preserved across filtering when they needed to be rebuilt from the filtered graph's new indices.

So the fix is not about filtering existing, it's about keeping the side maps consistent with the filtered graph after `#2469`.

milesj:
Yeah that makes sense. I thought I accounted for this. I think I assumed the indexes in `filter_map` were the new indexes, but they must be the old. Let me double check.

Edit: Yeah its the old index, whoops!

PR #2496, opened 2026-04-20
fix: avoid dispatcher stalls on sync-heavy action graphs

gabrielbryk:
## Summary
After upgrading to `2.2.2` in another project that uses moon, CI started stalling immediately after `SyncWorkspace` on a sync-heavy target with a single trailing noop task. I traced that back to dispatcher traversal in `JobDispatcher` and fixed it here.

This PR:
- memoizes nodes traversed during a dispatch pass so blocked sync-heavy subgraphs are not re-walked repeatedly
- adds dispatcher regression coverage for dense sync-heavy graphs
- adds a higher-level CLI regression that builds a dense sync-heavy workspace and verifies `moon exec app:noop` completes

## Root cause
`JobDispatcher::next()` recursively walks dependency chains to find the next runnable action.

On sync-heavy graphs with many shared `SyncProject` dependencies, that recursive search could repeatedly revisit the same blocked subgraphs while an early prerequisite was still the only runnable action. That left the dispatcher burning CPU after `SyncWorkspace` completed, and the pipeline appeared stalled even though the action graph itself was valid.

## What changed
- add per-pass traversal memoization in `JobDispatcher::find_applicable_index()`
- add dispatcher tests that exercise dense sync-heavy graphs directly
- add a CLI regression that builds a dense sync-heavy workspace dynamically and verifies both the graph shape and successful execution of `app:noop`

## Verification
- `cargo fmt --check`
- `cargo test -p moon_action_pipeline job_dispatcher::tests -- --nocapture`
- `cargo test -p moon_action_pipeline action_pipeline::priority::runs_priority_in_order -- --nocapture`
- `cargo test -p moon_cli exec::dispatcher_regression::runs_sync_heavy_noop_graph_without_stalling -- --nocapture`

### DIFF
--- a/crates/action-pipeline/src/job_dispatcher.rs
+++ b/crates/action-pipeline/src/job_dispatcher.rs
@@ -38,8 +38,9 @@ impl<'graph> JobDispatcher<'graph> {
         group: u8,
         index: NodeIndex,
         completed: &FxHashSet<NodeIndex>,
+        traversed: &mut FxHashSet<NodeIndex>,
     ) -> Option<NodeIndex> {
-        if self.visited.contains(&index) || completed.contains(&index) {
+        if !traversed.insert(index) || self.visited.contains(&index) || completed.contains(&index) {
             return None;
         }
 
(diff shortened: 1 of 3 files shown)


gabrielbryk:
@milesj please take a look at this when you get the chance, this bug is currently preventing my team from upgrading to latest version. We have a very large project graph and this bug causes CI to lock everytime

milesj:
@gabrielbryk Sure, give me an hour. This looks fine.

PR #2502, opened 2026-04-24
Add OpenTelemetry instrumentation for moon runs

gabrielbryk:
## Summary

Adds OpenTelemetry support for `moon` CLI runs, including traces, metrics, and optional logs.

This builds on the `starbase` OTEL support and instruments task execution, cache behavior, action graph/pipeline work, and process spawning with low-cardinality attributes suitable for collectors and dashboards.

## Dependency

Depends on moonrepo/starbase#186.

## Changes

- Add `--otel`, `--otel-logs`, and `--otel-service-name` CLI support.
- Add env-backed configuration through `MOON_OTEL`, `MOON_OTEL_LOGS`, and `MOON_OTEL_SERVICE_NAME`.
- Instrument task runs, execution attempts, cache lookup/hydration/archive, hash generation, action graph construction, pipeline concurrency waits, and process spawning.
- Add OTEL metrics for task runs, failures, retries, durations, cache lookups/hits, hydration, archive, execution, and concurrency wait.
- Keep OTEL dependencies feature-gated so `--no-default-features` does not pull OpenTelemetry.
- Avoid high-cardinality span attributes such as hashes, command lines, root paths, and full task commands.
- Move local-bin handoff before tracing initialization so Unix `exec` does not bypass tracing shutdown.
- Add OTEL docs and smoke scripts for OpenTelemetry Collector and Grafana Alloy + Loki.
- Add CLI and task-runner OTLP tests, including env-var configuration and cache-hit coverage.

## Verification

- `cargo fmt --check`
- `cargo check -p moon_task_runner --no-default-features`
- `cargo check -p moon_action_pipeline --no-default-features`
- `cargo check -p moon_cli --no-default-features`
- `cargo test -p moon_task_runner --test otel_test`
- `cargo test -p moon_cli --test otel_test`
- `scripts/otel/smoke-collector.sh`
- `scripts/otel/smoke-alloy-loki.sh`

### DIFF
--- a/crates/action-graph/src/action_graph_builder.rs
+++ b/crates/action-graph/src/action_graph_builder.rs
@@ -299,7 +299,13 @@ impl<'query> ActionGraphBuilder<'query> {
         Ok(())
     }
 
-    #[instrument(skip(self))]
+    #[instrument(
+        skip_all,
+        fields(
+            toolchain_id = %spec.id,
+            project_id = %project.id,
+        )
+    )]
     pub async fn install_dependencies(
         &mut self,
         spec: &ToolchainSpec,
@@ -393,7 +399,7 @@ impl<'query> ActionGraphBuilder<'query> {
         Ok(setup_toolchain_index)
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(project_id = %project.id))]
     pub async fn install_dependencies_by_project(
         &mut self,
         project: &Project,
@@ -402,7 +408,13 @@ impl<'query> ActionGraphBuilder<'query> {
             .await
     }
 
-    #[instrument(skip(self))]
+    #[instrument(
+        skip_all,
+        fields(
+            project_id = %project.id,
+            toolchain_count = toolchains.len(),
+        )
+    )]
     pub async fn install_dependencies_by_toolchains(
         &mut self,
         project: &Project,
@@ -419,7 +431,7 @@ impl<'query> ActionGraphBuilder<'query> {
         Ok(indexes)
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(task_target = %task.target))]
     pub async fn run_task(
         &mut self,
         task: &Task,
@@ -631,7 +643,7 @@ impl<'query> ActionGraphBuilder<'query> {
         }
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(task_target = %task.target))]
     pub async fn run_task_dependencies(
         &mut self,
         task: &Task,
@@ -676,7 +688,7 @@ impl<'query> ActionGraphBuilder<'query> {
         Ok(indexes)
     }
 
-    #[instrument(skip(self))]
+    #[instrument(skip_all, fields(task_target = %task.target))]
     pub async fn run_task_dependents(
         &mut self,
         task: &Task,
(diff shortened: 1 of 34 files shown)


crates/action-graph/src/action_graph_builder.rs
```diff
@@ -299,7 +299,13 @@ impl<'query> ActionGraphBuilder<'query> {
         Ok(())
     }
 
-    #[instrument(skip(self))]
```
milesj, line 302: I don't agree with changing any of the instrumentations. We lose a lot of data in non-otel contexts.

There are better ways to handle this, which can be done in a follow-up.

Undo all these changes so that the PR is very focused.

crates/action-pipeline/src/metrics.rs
```diff
@@ -0,0 +1,37 @@
...
+}
+
+impl ActionPipelineMetrics {
+    fn new() -> Self {
+        let meter = global::meter("moon_action_pipeline");
+
+        Self {
+            task_concurrency_wait_duration: meter
+                .u64_histogram("moon.task.concurrency_wait.duration")
+                .with_description("Moon task concurrency wait duration.")
```
milesj, line 21: moon is always lowercase.

crates/action-pipeline/Cargo.toml
```diff
@@ -5,11 +5,15 @@ edition = "2024"
 license = "MIT"
 publish = false
 
+[features]
```
milesj, line 8: Features aren't necessary in moon since these crates arent published.

crates/actions/Cargo.toml
```diff
@@ -4,6 +4,10 @@ version = "0.0.1"
 edition = "2024"
 publish = false
 
+[features]
```
milesj, line 7: Remove all the feautres in this PR.

crates/cli/src/shared.rs
```diff
@@ -35,7 +37,15 @@ fn get_version() -> String {
 fn get_tracing_modules() -> Vec<String> {
     let bag = GlobalEnvBag::instance();
     let mut modules = string_vec![
-        "moon", "proto", // "schematic",
+        "moon",
```
milesj, line 40: `moon` is all that's necessary. it should match everything.

crates/cli/Cargo.toml
```diff
@@ -62,11 +66,14 @@ moon_test_utils = { path = "../../legacy/core/test-utils" }
...
 moon_vcs = { path = "../vcs" }
 httpmock = "0.8.3"
+opentelemetry-proto = "0.31.0"
 proto_core = { workspace = true }
 rustc-hash = { workspace = true }
 serial_test = { workspace = true }
 starbase_archive = { workspace = true }
 starbase_sandbox = { workspace = true }
+tokio-stream = { version = "0.1.18", features = ["net"] }
+tonic = "0.14.5"
```
milesj, line 76: Use the workspace version

crates/process/src/exec_command.rs
```diff
@@ -37,21 +37,35 @@ impl Command {
                 .stdout(Stdio::piped())
                 .stderr(Stdio::piped());
 
-            let mut child = command.spawn().map_err(|error| ProcessError::Capture {
-                bin: self.get_bin_name(),
-                error: Box::new(error),
-            })?;
+            let spawn_span = self.create_spawn_span();
```
milesj, line 40: This is a lot of repeated same code. Can maybe just do something like `self.spawn(child)`? And then all this logic will be in that method.

crates/task-runner/src/labels.rs
```diff
@@ -0,0 +1,15 @@
+use moon_action::ActionStatus;
+
+pub(crate) fn action_status_label(status: ActionStatus) -> &'static str {
```
milesj, line 3: Add a `Display` impl to the enum instead.

crates/task-runner/src/metrics.rs
```diff
@@ -0,0 +1,235 @@
...
+    task_retry_total: Counter<u64>,
+    task_run_duration: Histogram<u64>,
+    task_hash_generation_duration: Histogram<u64>,
+    task_cache_lookup_total: Counter<u64>,
+    task_cache_hit_total: Counter<u64>,
+    task_cache_lookup_duration: Histogram<u64>,
+    task_hydration_duration: Histogram<u64>,
+    task_archive_duration: Histogram<u64>,
+    task_execution_duration: Histogram<u64>,
+    task_execution_attempts: Histogram<u64>,
```
milesj, line 23: Should attempts be a counter?

milesj (changes requested):

harlequin:
@gabrielbryk Can you please take also the MOON_TRACE_ID into consideration when creating the parent span?

https://github.com/moonrepo/moon/blob/3f13004ef639622c25690f3c8e747d06cd712343/crates/notifier/src/webhooks.rs#L82

Because with this we would have the possibility to group the distributed runs of moon easily.

PR #2504, opened 2026-04-27
Expose task content hash to task processes

gabrielbryk:
## Summary

Expose `MOON_TASK_CONTENT_HASH` to task child processes.

This hash uses Moon’s task-local and toolchain hash material, but excludes upstream dependency task result hashes. Moon’s existing cache behavior remains unchanged and continues to use `MOON_TASK_HASH` for hydrate/archive/remote cache behavior.

This gives external command wrappers a Moon-native static key for runtime-observed cache experiments without requiring them to reimplement Moon’s task hashing logic.

## Changes

- Generate both the full task hash and task content hash during task hash generation
- Preserve dependency edge identity in the content hash while excluding dependency result hash values
- Inject `MOON_TASK_CONTENT_HASH` alongside `MOON_TASK_HASH`
- Add tests for env injection and dep-hash exclusion behavior
- Document the new env var in cache docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+#### 🐞 Fixes
+
+- Fixed an issue where proto's `auto-clean` would remove tools installed by moon as they weren't
+  marked as used.
+
 ## 2.2.3
 
 #### 🐞 Fixes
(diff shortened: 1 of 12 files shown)


milesj:
Going to land this in 2.3 https://github.com/moonrepo/moon/pull/2516

We'll see if this kind of PR is still necessary.

PR #2505, opened 2026-04-27
fix(cli): clear inherited O_NONBLOCK on stdio at startup

gabrielbryk:
## Summary

Closes the EAGAIN variant of #2465.

When moon is invoked from an environment that hands children file descriptors with `O_NONBLOCK` set on the underlying open file description — most notably the GitHub Actions log forwarder — Rust's `write_all` returns `WouldBlock` mid-task and the binary exits 1 with `Error: console::write_failed` / `Resource temporarily unavailable (os error 11)`. The original bug report observed this in ~13% of CI invocations across many unrelated branches.

This patch adds a tiny unix-only `normalize_stdio_blocking()` that runs once at the top of `run_cli` (before any output) and clears `O_NONBLOCK` from fds 0, 1, 2 if it's set. After that, every writer in the process — `println!`, the panic hook, `tracing`, `miette`, `clap_complete`, `starbase_console`, third-party crates — sees blocking stdio and can use `write_all` correctly. SIGPIPE handling at `crates/cli/src/shared.rs:85` (`sigpipe::reset()`) already covered the EPIPE side of #2465; no change needed there.

## Why a process-wide clear, not per-writer retry

The starbase console writer was the most visible failure site, but it isn't the only one. While validating this fix, removing `normalize_stdio_blocking()` and replaying the same workload (`moon completions --shell bash` piped to a slow reader) crashed inside `clap_complete::generate`'s writer instead — different crate, same EAGAIN, same fatal exit. Auditing every `write_all` in moon's transitive dep graph isn't tractable; clearing the OS-level flag once is.

Trade-off: `fcntl(F_SETFL)` modifies the open file description, which is shared with the parent's pipe. In practice `cargo`, `rustc`, and coreutils all rely on this being safe. The parent set `O_NONBLOCK` incorrectly for a child that doesn't expect it; restoring blocking semantics matches what stdlib assumes.

## Tests

`crates/cli/src/stdio.rs` adds 4 regression tests:

- `clears_o_nonblock_when_set` — opens a real pipe via `libc::pipe`, sets `O_NONBLOCK` on the write end, calls the helper, asserts the flag is cleared.
- `leaves_blocking_fd_alone` — idempotency on already-blocking fds.
- `bad_fd_is_silent` — graceful handling of invalid fds (no panic).
- `normalize_stdio_blocking_clears_inherited_nonblock` — public-API smoke test asserting fds 0/1/2 are blocking after the call.

`libc` is added under `[target.'cfg(unix)'.dependencies]` so Windows builds are unaffected.

## Validation against the original bug

End-to-end repro using a wrapper that flips stdout to `O_NONBLOCK` and execs moon — exact GHA-style condition with `moon completions --shell bash | (sleep 1; cat > /dev/null)`:

| Build | Exit | Stderr |
|---|---|---|
| **Without this patch** | 1 | `Error: ... os error 11, kind: WouldBlock` (panicked in `clap_complete`) |
| **With this patch** | 0 | (empty) |

A companion library-side fix is open at [moonrepo/starbase#187](https://github.com/moonrepo/starbase/pull/187). It is not required for this PR — clearing `O_NONBLOCK` here neutralizes EAGAIN for every writer in the process, including third-party crates.

## Test plan
- [x] `cargo test -p moon_cli --bin moon stdio::` — 4/4 pass
- [x] `cargo build -p moon_cli --bin moon` — clean (no new warnings)
- [x] `cargo clippy -p moon_cli --bin moon --no-deps` — clean
- [x] End-to-end repro: patched moon survives `O_NONBLOCK` stdout + slow reader; unpatched moon dies with `os error 11`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5194,6 +5194,7 @@ dependencies = [
  "clap",
  "extism",
  "httpmock",
+ "libc",
  "liblzma",
  "mimalloc",
  "moon_affected",
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -51,6 +51,9 @@ wiggle = "~41.0.4"
 # Overrides
 liblzma = { version = "0.4.6", features = ["static"] }
 
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
+
 [dev-dependencies]
 moon_affected = { path = "../affected" }
 moon_cache = { path = "../cache" }
--- a/crates/cli/src/main.rs
+++ b/crates/cli/src/main.rs
@@ -1,5 +1,6 @@
 mod lookup;
 mod shared;
+mod stdio;
 
 use starbase::MainResult;
 use std::env;
--- a/crates/cli/src/shared.rs
+++ b/crates/cli/src/shared.rs
@@ -82,6 +82,7 @@ fn exec_local_bin(mut command: Command) -> std::io::Result<u8> {
 }
 
 pub async fn run_cli(args: Vec<OsString>) -> MainResult {
+    crate::stdio::normalize_stdio_blocking();
     sigpipe::reset();
 
     // Detect info about the current process
--- a/crates/cli/src/stdio.rs
+++ b/crates/cli/src/stdio.rs
@@ -19,11 +19,7 @@
 pub fn normalize_stdio_blocking() {
     use std::os::raw::c_int;
 
-    for fd in [
-        libc::STDIN_FILENO,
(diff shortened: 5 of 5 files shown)


milesj:
@gabrielbryk Looks like theres a format issue.

milesj:
@gabrielbryk Release a patch for `starbase_console` if you want to pull that fix in here as well.

milesj:
@gabrielbryk Can you get the build passing.

PR #2512, opened 2026-04-30
fix: skip uninitialized submodules when loading the git tree

### LINKED ISSUE #2510: [bug] moon v2 fails with "Failed to execute git and capture output" when a submodule path is not checked out
## Describe the bug

After upgrading to moon 2.x, any `moon run` fails immediately if `.gitmodules` references a submodule whose path doesn't exist on disk (e.g. submodule has `update = none` and was never initialized). The error is:

```
Error: process::capture::failed

  × Failed to execute git and capture output.
  ╰─▶ No such file or directory (os error 2)
```

This appears to be a regression of #1954 (fixed in moon v1's `gitx`) re-introduced by the Git v2 rewrite in #2184. moon iterates each submodule listed in `.gitmodules` and tries to spawn git with `cwd = <submodule path>`; when the path doesn't exist, `posix_spawn` returns `ENOENT` before the command can run.

## Steps to reproduce

````bash
mkdir /tmp/moon-submodule-repro && cd /tmp/moon-submodule-repro
git init -q
git config user.email t@t && git config user.name t

cat > .gitmodules <<EOF
[submodule "missing"]
	path = vendor/missing
	url = https://github.com/example/example.git
	update = none
EOF

mkdir -p .moon && cat > .moon/workspace.yml <<EOF
$schema: 'https://moonrepo.dev/schemas/workspace.json'
projects:
  - 'app'
EOF

mkdir -p app && cat > app/moon.yml <<EOF
$schema: 'https://moonrepo.dev/schemas/project.json'
layer: application
tasks:
  hello:
    command: 'echo'
    args: ['hi']
EOF

git add -A && git commit -q -m init
moon run app:hello   # fails with the error above
````

Workaround that confirms the cause: `mkdir -p vendor/missing` and re-run — task succeeds.

## Trace (relevant lines)

```
[DEBUG] moon_vcs::git::git_client  Loading submodules from .gitmodules  modules_file="/private/tmp/moon-submodule-repro/.gitmodules"
[DEBUG] moon_process::exec_command  Running command git status --porcelain --untracked-files --ignore-submodules -z  cwd="/private/tmp/moon-submodule-repro"
[TRACE] starbase::app  Error: Failed to execute git and capture output.
Error: process::capture::failed
```

The main-tree `git status` is the last logged command; the per-submodule spawn fails at `posix_spawn` so it never reaches the trace's "Running command" log.

## Expected behavior

moon should either skip submodules whose `path` doesn't exist on disk (as v1 did after #1954), or surface a clearer error pointing at `.gitmodules`. Submodules with `update = none` are a legitimate setup — typically large optional vendor submodules that aren't required for most tasks.

## Environment

- moon: 2.2.3
- OS: macOS 15 (Darwin 25.0.0 arm64)
- git: 2.50.1 (Apple Git-155)

### PR DESCRIPTION
**Pull request by xiang-baseten**
## Summary
After upgrading to moon v2, any `moon run` fails immediately if `.gitmodules` references a submodule whose path hasn't been checked out (e.g. `update = none`, or a clone without `--recurse-submodules`). The error surfaces as the misleading:

```
Error: process::capture::failed
  × Failed to execute git and capture output.
  ╰─▶ No such file or directory (os error 2)
```

This re-applies the v1 fix from #1955 that was lost in the Git v2 rewrite (#2184).

Fixes #2510

## Root cause
`Git::load` in `crates/vcs/src/git/git_client.rs` iterates every submodule listed in `.gitmodules` and adds it to `git.submodules` without checking whether the path is actually checked out on disk. Subsequent fan-out commands (`get_changed_files`, `get_changed_files_between_revisions`, `get_file_tree`) spawn `git` with `cwd = <submodule work_dir>`. When the directory doesn't exist, `posix_spawn` returns `ENOENT` before the command can run, which is why the trace's `Running command` log never fires for the failing spawn.

The v1 implementation in `crates/vcs/src/gitx/tree.rs` filtered submodules whose `<work_dir>/.git` was missing. That check wasn't carried over in #2184.

## What changed
- `crates/vcs/src/git/git_client.rs`: filter submodules at load time on `<work_dir>/.git` existing, with a debug log when one is skipped. Matches the v1 behavior verbatim.
- `crates/vcs/tests/git_test.rs`:
  - `git::submodules::doesnt_error_if_submodules_arent_checked_out` was inadvertently asserting the buggy behavior (`!is_empty()`); flipped to `is_empty()` and added a `get_changed_files().await.unwrap()` call so the test exercises the fan-out spawn path that was failing.
  - `git::root::loads_trees_when_bare`: bare clones with `--recurse-submodules --bare` don't materialize submodule worktrees on disk, so `submodules` is now empty after the filter. Updated the assertion accordingly.
  - `git::worktree::loads_trees{,_when_bare}`: unchanged — those sandboxes run `git submodule update`, so the work_dirs do exist.
- `CHANGELOG.md`: Unreleased entry.

## Verification
- `cargo test -p moon_vcs --test git_test 'submodules' -- --test-threads=1`
- `cargo test -p moon_vcs --test git_test 'loads_trees' -- --test-threads=1`

Both pass. The repro from #2510 (synthetic repo with `update = none` submodule) succeeds with this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 - Fixed an issue where proto's `auto-clean` would remove tools installed by moon as they weren't
   marked as used.
+- Fixed a regression where moon would fail with `Failed to execute git and capture output` when
+  `.gitmodules` referenced a submodule that hadn't been checked out (e.g. `update = none`).
+  Uninitialized submodules are now skipped, matching the v1 behavior.
 
 ## 2.2.3
 
--- a/crates/vcs/src/git/git_client.rs
+++ b/crates/vcs/src/git/git_client.rs
@@ -143,12 +143,27 @@ impl Git {
                         if let Ok(work_dir) = sub.work_dir()
                             && let Ok(rel_path) = sub.path()
                         {
+                            let work_dir = clean_components(if work_dir.is_absolute() {
+                                work_dir
+                            } else {
+                                repository_root.join(work_dir)
+                            });
+
+                            // Ensure the submodule has actually been checked out.
+                            // When a submodule is declared in `.gitmodules` but never
+                            // initialized (e.g. `update = none`), running git in its
+                            // missing work_dir would fail with ENOENT.
+                            if !work_dir.join(".git").exists() {
+                                debug!(
+                                    submodule = ?work_dir,
+                                    "Encountered a submodule that hasn't been checked out, skipping it"
+                                );
+
+                                continue;
+                            }
+
                             submodules.push(GitTree {
-                                work_dir: clean_components(if work_dir.is_absolute() {
-                                    work_dir
-                                } else {
-                                    repository_root.join(work_dir)
-                                }),
+                                work_dir,
                                 git_dir: clean_components(sub.git_dir()),
                                 type_of: GitTreeType::Submodule,
                                 path: RelativePathBuf::from(rel_path.to_string()),
--- a/crates/vcs/tests/git_test.rs
+++ b/crates/vcs/tests/git_test.rs
@@ -169,25 +169,10 @@ mod git {
             assert_eq!(git.worktree.work_dir, sandbox.path());
             assert_eq!(git.worktree.path.as_str(), "");
             assert_eq!(git.worktree.type_of, GitTreeType::Root);
-            assert_eq!(
-                git.submodules,
-                vec![
-                    GitTree {
-                        git_dir: sandbox.path().join("modules/submodules/mono"),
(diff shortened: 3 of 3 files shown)


milesj:
@xiang-baseten I think an error may be better as projects/tasks may be referenced in this submodule, and users may be confused why things aren't working.

They error should probably only happen when that submodule is needed, and not on load.

milesj:
Thought about this more, and detecting when a submodule is in use is actually quite difficult right now, so this is ok.

nicolasgere:
Thank you!

PR #2513, opened 2026-04-30
fix: Query changed-files now respects --head and avoids local status mixing

### LINKED ISSUE #2511: [bug] moon query changed-files ignores --head and mixes in local working tree / git status
**Describe the bug**

  `moon query changed-files --base <X> --head <Y>` does not return the diff between `X` and `Y`. It silently drops `--head` and instead returns `git diff <X>` (one-arg, i.e. base → working tree) **unioned with `git status`** (local uncommitted state). Files touched in commits *after* `--head`, plus any local
  modifications, show up as false positives. This breaks any commit-by-commit walk (release notes, per-commit "which projects were affected" gating, etc.).

  **Steps to reproduce**

  Repro repo: https://github.com/nicolasgere/moon-bug-repro — three linear commits: `PARENT` (root), `MID` (touches only `pkg-a/a.txt`), `TIP` (touches only `pkg-b/b.txt`), plus later README commits on `main`.

  1. Clone and install the pinned Moon version:
     ```bash
     git clone https://github.com/nicolasgere/moon-bug-repro.git
     cd moon-bug-repro
     proto install   # pins moon 2.2.3 via .prototools
     ```
  2. Resolve the three commits:
     ```bash
     MID="$(git rev-list -1 --grep='^commit2: only pkg-a' HEAD)"
     PARENT="$(git rev-parse "${MID}^")"
     ```
  3. Compare Git vs Moon for the `PARENT..MID` range:
     ```bash
     git diff --name-only "$PARENT" "$MID"
     # pkg-a/a.txt

     moon query changed-files --base "$PARENT" --head "$MID" -q | jq '.files'
     # [
     #   "pkg-b/b.txt",   <- false positive (only changed in TIP, after --head)
     #   "pkg-a/a.txt",
     #   "README.md"      <- false positive (only changed on main after TIP)
     # ]
     ```
  4. `--remote` produces the same wrong result.
  5. With `MOON_LOG=trace`, Moon parses the args correctly:
     ```
     moon_app::queries::changed_files  Against remote using base "<PARENT>" with head "<MID>"
     ```
     …but then issues a one-arg `git diff` (no `<HEAD>`):
     ```
     git diff --name-status --no-color --relative --ignore-submodules -z <PARENT>
     ```
     …followed by:
     ```
     moon_app::queries::changed_files  Against local index
     git status --porcelain --untracked-files --ignore-submodules -z
     ```
     So the effective query is `git diff <BASE>` ∪ `git status`, regardless of `--head`. Confirmed by running the same one-arg command directly — it returns exactly Moon's buggy file list.

  **Expected behavior**

  When both `--base` and `--head` are passed as explicit revisions, `moon query changed-files` should:

  1. Run `git diff --name-status … <BASE> <HEAD>` (two-arg) and return only the files in that tree-to-tree diff.
  2. Not consult `git status` / the local index — the user requested a diff between two SHAs, not a working-tree comparison.

  i.e. the result should match `git diff --name-only <BASE> <HEAD>`.

  **Screenshots**

  N/A — terminal output included above.

  **Environment**

  ```
    System:
      OS: macOS 26.3
      CPU: (18) arm64 Apple M5 Pro
      Memory: 461.39 MB / 48.00 GB
      Shell: 5.9 - /bin/zsh
    Binaries:
      Node: 22.14.0 - /Users/nicolas/.proto/tools/node/22.14.0/bin/node
      npm: 10.9.2 - /Users/nicolas/.proto/tools/node/22.14.0/bin/npm
      pnpm: 10.29.3 - /Users/nicolas/.proto/shims/pnpm
    Utilities:
      Git: 2.53.0 - /opt/homebrew/bin/git

  moon: 2.2.3 (installed via proto 0.56.4)
  ```

  **Additional context**

  - Affects any workflow that walks history commit-by-commit and pipes ranges into `moon query changed-files` or `moon query projects --affected` — release-notes generators and per-commit affected-project gating will silently report wrong results.
  - The bug shows up even with a clean working tree (no uncommitted edits); the false positives come from commits *after* `--head`, because `git diff <BASE>` compares base → working tree.
  - Suspected location: the changed-files query path in `moon_app::queries::changed_files` — the explicit-`--head` branch appears to fall through to the same code as the implicit/local case.

nicolasgere:
## Summary
Fixes #2511.

When `--base` and `--head` were both passed to `moon query changed-files`, the `--head` revision was silently ignored and the command fell back to a one-argument `git diff <base>` (base → working tree). It also unconditionally merged `git status` output, so commits after `--head` and local uncommitted changes appeared as false positives.

## Changes
- **git_client.rs:** Pass `head_revision` into `exec_diff` instead of an empty string, so `git diff` runs with two explicit revisions.
- **changed_files.rs:** Only merge local `git status` changes when `head` is not explicitly set. When the user asks for a diff between two SHAs, we return a pure tree-to-tree diff.

## Verification
- `moon query changed-files --base <X> --head <Y>` now returns the same files as `git diff --name-only <X> <Y>`.
- Open-ended queries (no `--head`) still include local working tree changes, preserving existing behavior.

### DIFF
--- a/crates/vcs/src/git/git_client.rs
+++ b/crates/vcs/src/git/git_client.rs
@@ -530,7 +530,7 @@ impl Vcs for Git {
             .unwrap_or(base_revision);
 
         // Load from root repo
-        changed_files.merge(self.worktree.exec_diff(merge_base_revision, "").await?);
+        changed_files.merge(self.worktree.exec_diff(merge_base_revision, head_revision).await?);
 
         // Load from each submodule
         if !self.submodules.is_empty() {
--- a/crates/app/src/queries/changed_files.rs
+++ b/crates/app/src/queries/changed_files.rs
@@ -133,10 +133,12 @@ async fn query_changed_files_without_stdin(
         }
     }
 
-    // Always include local changes
-    trace!("Against local index");
+    // Only include local changes when not explicitly comparing two revisions
+    if head_value.is_none() || options.local {
+        trace!("Against local index");
 
-    changed_files_map.merge(vcs.get_changed_files().await?);
+        changed_files_map.merge(vcs.get_changed_files().await?);
+    }
 
     if options.status.is_empty() {
         debug!(


crates/vcs/src/git/git_client.rs
```diff
@@ -530,7 +530,7 @@ impl Vcs for Git {
             .unwrap_or(base_revision);
 
         // Load from root repo
-        changed_files.merge(self.worktree.exec_diff(merge_base_revision, "").await?);
+        changed_files.merge(self.worktree.exec_diff(merge_base_revision, head_revision).await?);
```
milesj, line 533: I can't remember why it is this way.

But I think `head_revision` should only be used if `merge_base_revision` is the same as the original `base_revision`, otherwise the new merge base diverges.

PR #2516, opened 2026-05-02
feat: add cacheStrategy field to task dependencies

ateirney-nz:
Adds a per-dep `cacheStrategy` field that controls how a dependency's change affects the consumer task's hash.

Three strategies:

- `hash` — the dep's hash is mixed into the consumer's hash; any dep change invalidates.
- `ignored` — the dep is a sequencing edge only; its changes never invalidate the consumer.
- `outputs` — only the dep's *outputs* contribute, not its hash. The dep's output files are injected as inputs of the consumer task, so changes to upstream sources that don't actually change build artifacts no longer invalidate downstream caches.

The most common use case is breaking input chaining in build pipelines: rebuild only when an upstream's outputs change, not when its inputs change.

The field is optional. When omitted, the effective default is resolved during task expansion from the dep itself: `hash` if the dep declares outputs, `ignored` if it doesn't. **Behavior change:** tasks that depend on output-less deps (e.g. `lint`, `test`) will see fewer cache invalidations than before. Set `cacheStrategy: 'hash'` explicitly to restore the previous behavior for a given dep.

A new `TaskLookup` trait in `task-expander` exposes a dep's expanded outputs without taking a hard dependency on `task-graph` (which would create a cycle). `TaskGraph` implements it; expansion is memoized.


crates/config/src/task_config.rs
```diff
@@ -138,6 +161,11 @@ config_struct!(
         /// @since 1.20.0
         #[serde(default, skip_serializing_if = "Option::is_none")]
         pub optional: Option<bool>,
+
+        /// Controls how this dependency invalidates the current task's cache.
+        /// @since 2.2.4
```
milesj: Change to 2.3.0

crates/task-expander/src/task_expander.rs
```diff
@@ -55,13 +61,77 @@ impl<'graph> TaskExpander<'graph> {
...
+        }
+
+        Ok(())
+    }
+
+    // Expands a dep's outputs using a token expander bound to the dep's own
+    // project — tokens like $projectRoot or @group(...) must resolve against
+    // the producing project, not the consumer's.
+    fn expand_outputs_for_dep_target(&self, target: &Target) -> miette::Result<ExpandedResult> {
+        let project = self
```
milesj: It will need the expanded project so that tokens/etc are resolved.

crates/task-expander/src/task_expander.rs
```diff
@@ -55,13 +61,77 @@ impl<'graph> TaskExpander<'graph> {
...
+            .project_graph
+            .get_unexpanded(target.get_project_id()?)?;
+        let task = self.task_lookup.get_unexpanded(target)?;
+        let mut token = TokenExpander::new(self.project_graph, project, self.context);
```
milesj: Token expander already exists at `this.token`

crates/task-expander/src/task_expander.rs
```diff
@@ -55,13 +61,77 @@ impl<'graph> TaskExpander<'graph> {
...
+    }
+
+    // Expands a dep's outputs using a token expander bound to the dep's own
+    // project — tokens like $projectRoot or @group(...) must resolve against
+    // the producing project, not the consumer's.
+    fn expand_outputs_for_dep_target(&self, target: &Target) -> miette::Result<ExpandedResult> {
+        let project = self
+            .project_graph
+            .get_unexpanded(target.get_project_id()?)?;
+        let task = self.task_lookup.get_unexpanded(target)?;
```
milesj: This also needs to be the expanded task.

crates/config/src/task_config.rs
```diff
@@ -121,6 +121,29 @@ config_unit_enum!(
     }
 );
 
+config_unit_enum!(
+    /// Controls how a task dependency invalidates the current task's cache entry.
+    /// @since 2.2.4
+    #[derive(ConfigEnum)]
+    pub enum TaskDependencyCacheStrategy {
+        /// Use the dependency task's hash to invalidate the current task's cache.
+        /// This is the default behavior and will invalidate whenever the dependency changes.
```
milesj: I think it would be nice if the default is dependent on the dep task.

Has outputs = hash
No outputs = ignored

milesj (changes requested):
Good start, a few comments.

This will also land in v2.3, so update the base branch to `develop-2.3`.

crates/config/src/task_config.rs
ateirney-nz: I have implemented this, and adjusted the documentation and that accordingly.  I have noted there is a behavioral change with this.

crates/config/src/task_config.rs
ateirney-nz: All references to 2.2.4 have been changed to 2.3.0

crates/task-expander/src/task_expander.rs
ateirney-nz: Updated the `TaskLookup` trait to expose `get_outputs`, returning the dep's *expanded* outputs directly — `TaskGraph` triggers a memoized expansion of the dep on lookup. That meant some changes in `TaskGraph::internal_get` to release the entry guard during expansion to avoid a deadlock when the dep itself fans out to other deps.

crates/task-expander/src/task_expander.rs
ateirney-nz: Addressed via the `TaskLookup` change below — outputs are now read off the already-expanded dep, so the consumer-side expander never needs to look up a project here.

crates/task-expander/src/task_expander.rs
ateirney-nz: Addressed via the `TaskLookup` change below — outputs come back already token-expanded, so neither `self.token` nor a new `TokenExpander` is needed at this site.

crates/task-expander/src/task_expander.rs
```diff
@@ -55,13 +59,92 @@ impl<'graph> TaskExpander<'graph> {
...
 
         Ok(task)
     }
 
+    // Fills in `cache_strategy` for deps the user left unset. The default
+    // depends on the dep itself: `Hash` when the dep declares outputs (changes
+    // there should invalidate the consumer), `Ignored` when it doesn't (the
+    // dep is a sequencing edge only). Explicit strategies are left alone.
+    #[instrument(skip_all)]
+    pub fn resolve_deps_cache_strategies(&mut self, task: &mut Task) -> miette::Result<()> {
```
milesj: I think this should probably be in the task deps builder, just to keep similar logical localized: https://github.com/moonrepo/moon/blob/master/crates/task-builder/src/task_deps_builder.rs

crates/task-graph/src/task_graph.rs
```diff
@@ -155,25 +155,39 @@ impl TaskGraph {
...
-        let task = match self.tasks.entry_sync(target.to_owned()) {
-            Entry::Occupied(entry) => Arc::clone(entry.get()),
-            Entry::Vacant(entry) => {
-                let expander = TaskExpander::new(
-                    &self.project_graph,
-                    self.project_graph
-                        .get_unexpanded(target.get_project_id()?)?,
-                    &self.context,
-                );
+        if let Entry::Occupied(entry) = self.tasks.entry_sync(target.to_owned()) {
```
milesj: What's the reason for this change?

This breaks async lock guarantees, since task expansion now happens outside of a lock. Theoretically, this means that the same task can be expanded multiple times in parallel.

crates/task-graph/src/task_graph.rs
```diff
@@ -155,25 +155,39 @@ impl TaskGraph {
...
+        })
+    }
+}
 
-        Ok(task)
+impl TaskLookup for TaskGraph {
+    fn get_outputs(&self, target: &Target) -> miette::Result<TaskOutputs> {
+        let task = self.internal_get(target)?;
+        Ok(TaskOutputs {
+            files: task.output_files.clone(),
```
milesj: Is it possible to use lifetimes/refs here to avoid cloning?

milesj (commented):
Just a few follow-up Qs.

crates/task-expander/src/task_expander.rs
ateirney-nz: Done. The default-strategy resolution is now in `check_and_create_dep` inside
`task_deps_builder.rs`, driven by `query_task_has_outputs` on the `TasksQuerent` trait.
The expander no longer has any strategy logic.

crates/task-graph/src/task_graph.rs
ateirney-nz: The `TaskLookup` trait no longer has `get_outputs` at all — it now exposes
`get_task(&Target) -> miette::Result<Arc<Task>>`. Consumers read `output_files` and
`output_globs` directly off the shared `Arc<Task>`, so there's no copy of the output maps.

crates/task-graph/src/task_graph.rs
ateirney-nz: The refactor was driven by a deadlock observed locally: `scc::HashMap::entry_sync` holds a
non-reentrant shard lock for the duration of the call. When task A is expanding and calls
`get_task(B)` (e.g. via `inject_deps_outputs`), `internal_get(B)` tries to acquire
`entry_sync(B)` — which deadlocks if A and B land in the same shard, since the lock from
A's `entry_sync` is still held.

The `OnceCell` fix addresses this by releasing the shard lock immediately after inserting
(or finding) the `Arc<OnceCell>` entry, then calling `get_or_try_init` outside the lock.
This also closes the double-expansion race you noted: concurrent callers on the same target
share one `Arc<OnceCell>` and `get_or_try_init` ensures only one expansion runs. I am not sure how you feel about using `OnceCell`.

milesj: It _should_ be fine. We'll see how performance goes.

milesj:
@ateirney-nz Code is good, just CI failures.

ateirney-nz:
@milesj  I just corrected the format and lint errors, that was my bad for not running the just targets beforehand.

Can you please trigger the associated workflows again to ensure I have not messed anything else up.

In relation to the performance degradation,  any pointers on how I should look to investigate those further?

ateirney-nz:
@milesj Is there anything else that you would want me to do for this pull request?

milesj:
@ateirney-nz Nope, looks good! Just been busy.

PR #2527, opened 2026-05-14
internal: Co-locate all hash/digest/fingerprint logic into the same crate.

milesj:


milesj:
@codspeedbot fix this regression

milesj:
@codspeedbot debug why a benchmark failed

PR #2555, opened 2026-06-04
Fix affected stdin terminal detection

DominicVonk:
## Summary

- Only auto-read changed files from stdin when stdin is not an interactive terminal.
- Adds a comment explaining that interactive terminals would otherwise hang waiting for EOF.
- Adds coverage for `moon query affected` consuming changed files from stdin.

## Validation

- `git diff --check`
- Could not run Rust formatting/tests locally because `cargo` and `rustfmt` are not installed in this container.

### DIFF
--- a/crates/app/src/queries/changed_files.rs
+++ b/crates/app/src/queries/changed_files.rs
@@ -225,7 +225,8 @@ pub async fn query_changed_files_for_affected(
     let mut options = QueryChangedFilesOptions {
         default_branch: ci,
         local: !ci,
-        stdin: true,
+        // Only auto-read stdin for piped input, otherwise interactive terminals hang waiting for EOF.
+        stdin: !stdin().is_terminal(),
         ..Default::default()
     };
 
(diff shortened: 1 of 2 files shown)


milesj:
@DominicVonk Unfortunately this is a breaking change.

dominicvonk-milence:
Sorry, currently still working on it

milesj:
@DominicVonk We already have checks around stdin here: https://github.com/moonrepo/moon/blob/master/crates/app/src/queries/changed_files.rs#L189

What issues are you running into exactly?

harlequin:
I have also the same problem, our environment is azure devops pipelines.
My workaround so far is 

```
set -xe     
faketty () {
  script -qefc "$(printf "%q " "$@")" /dev/null
}     
faketty moon query tasks --affected --upstream deep --downstream deep
```

When running without this faketty it will wait in an endless loop.

DominicVonk:
This exact problem our current workaround is to pipe /dev/null as stdin

milesj:
`stdin().is_terminal()` is not a good enough check then, as that is already in place.

DominicVonk:
Yes, indeed. I will debug further and update the PR

PR #2561, opened 2026-06-09
feat: add flake.nix for Linux development and Nix packaging

Flexksx:
## Summary

- Adds `flake.nix` exposing `packages.default` (the `moon` + `moonx` binaries) and `devShells.default` for Linux development
- Targets `x86_64-linux` and `aarch64-linux`
- Pins the exact Rust 1.96.0 toolchain via `rust-overlay` reading `rust-toolchain.toml`
- Adds `.github/workflows/nix.yml` as a guardrail — runs `nix flake check` and `nix build` on every relevant change

## Motivation

The [nixpkgs derivation for moon](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/mo/moon/package.nix) is currently at **1.41.8**, while the project is at **2.3.2**. A flake shipped with the repo lets Nix users pin directly to any commit or release without waiting for a nixpkgs update:

```bash
nix run github:moonrepo/moon -- --version

nix develop github:moonrepo/moon
```

## Technical notes

**`OPENSSL_NO_VENDOR = 1`** — `reqwest` enables `native-tls-vendored` for musl targets. On glibc Linux this forces pkg-config lookup of the system openssl instead of compiling it from source.

**`protobuf` in `nativeBuildInputs`** — `crates/daemon-proto/build.rs` invokes `protoc` via `tonic-prost-build` at compile time. This dependency was added after the nixpkgs derivation was last updated.

**`cargoLock.lockFile`** — the project has no git-sourced Cargo dependencies, so no `outputHashes` are needed.

**`doCheck = false`** — tests download Node, Bun, Deno, etc. at runtime and cannot pass in the Nix sandbox.

## Test plan

- [x] `nix flake check --no-build` passes locally
- [x] Derivation evaluates to a valid `.drv` on `x86_64-linux`
- [ ] `nix build .#default` succeeds in CI (workflow added in this PR)


milesj:
@Flexksx I'm not familiar with nix, what does this do exactly? Allow people to use moon in their nix environment?

Flexksx:
@milesj hi! 

> Allow people to use moon in their nix environment?

Yes, that's the biggest reason

Nix lets you package packages in a way that keeps them isolated, so I could have 2 different projects using `moon` and I could configure them to use slightly different versions of `moon`.

And each project would have its own isolated `moon` that is provided through the nix store.

I just recently stumbled across `moon` and I'm using it for a personal project and to be fair I'm amazed at how old is the current nixpkgs version for it, so I opened a PR as well https://github.com/NixOS/nixpkgs/pull/529832

So for my project, I had to configure nix in a way that would build it from a specific, newer commit in the repo.
Which does not really have that much ergonomics :shrug: 

So adding a `flake.nix` would let as well other people easier interact with it, install it on their machines.

Also having a `flake.nix` lets other people with nix systems develop and contribute safely, since normally on NixOS you don't really install any packages globally, everything is managed with nix

milesj:
@Flexksx Sorry for the late response. The code looks fine to me. Can you just add a quick `docs/` file on how to manager/update this going forward.

Flexksx:
Sorry for the late reply

@milesj added docs/NIX.md and some more lines in the `install.mdx`

Most of it is about what needs a bump and when, which is not really much.
The flake's package version is read from the `crates/cli/Cargo.toml` so a release does not need a flake edit.

The only thing is to run `nix flake update` whenever you want the pinned inputs moved forward, e.g. after a new nixpkgs release that has stuff you want.

Heads up that github does not run anything for this PR on your side, fork runs need a maintainer to approve them.

milesj:
Thanks!

PR #2571, opened 2026-06-15
feat(vcs): jj-aware git backend, incl. Jujutsu workspace support

lamalex:
Adds a `JjAwareGit` decorator that wraps the Git backend inside a Jujutsu workspace and routes working-copy/revision operations through jj.
- Detect via `jj root`.
- get_changed_files.
- get_local_branch → bookmark at @ or short change-id.
- get_local_branch_revision → @ commit id.
- setup_hooks → still installs for git users, warns jj won't run them.
- Everything else (file hashes/tree, repo root/slug) delegates to git.

Auto-selected in session.rs when JjAwareGit::is_jj_workspace(). Plain-git checkouts and committed config are unaffected.

### DIFF
--- a/crates/app/src/session.rs
+++ b/crates/app/src/session.rs
@@ -24,7 +24,7 @@ use moon_process::ProcessRegistry;
 use moon_project_graph::ProjectGraph;
 use moon_task_graph::TaskGraph;
 use moon_toolchain_plugin::*;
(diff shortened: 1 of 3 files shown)


milesj:
I'm going to defer this till maybe v2.5, because I really have to think about the side-effects of this.

PR #2602, opened 2026-07-07
fix: Fix workspace graph caching. Avoid unnecessary plugin calls.

milesj:


cedws:
Much love for this fix, it's made a huge difference.

PR #2622, opened 2026-07-16
fix: Comply with the RE API contract when uploading to a remote cache

KeyZer:
Remote cache uploads are rejected by backends that validate the Bazel Remote Execution contract, so the cache never fills and every task re-runs. I hit this pointing moon at [Depot Cache](https://depot.dev/docs/cache/integrations/moonrepo); on a real repo **6 of 6 tasks failed to upload**.

The failures are non-fatal warnings, so it looks like remote caching is simply doing nothing rather than erroring.

## Causes

**1. `UpdateActionResult` populates `stdout_raw`/`stderr_raw`.**

```
Failed to store cache manifest storage="grpc-remote-cache"
error="Client specified an invalid argument: client should not populate stdout_raw during upload"
```

The API reserves those fields for server responses. `collect_blob_inputs` already uploads stdout/stderr to the CAS and `into_bazel_action_result` references them by digest, so the inline copy was redundant as well as rejected.

**2. The action digest is never uploaded to the CAS.**

```
error="Some requested entity was not found: action digest <hash>/2585 not found in CAS"
```

`store_manifest` calls `UpdateActionResult` with an action digest the server cannot resolve. Conveniently, `ContentHasher::generate_hash` is `hash_bytes(serialize())` and `HashEngine::save_manifest` sets `size = data.len()` over those same bytes — so the action digest is exactly `(sha256(json), len(json))` of the hash manifest already written to `.moon/cache/hashes/<hash>.json`. That manifest *is* the blob the digest addresses, so uploading it alongside the outputs satisfies the check, with content-addressing correct by construction and no change to cache keying.

**3. (Latent, surfaced by fixing 1.) Raw stdio bytes were trusted whenever a digest was present.**

`from_bazel_action_result` set `stdout_bytes` from `stdout_raw` if a *digest* existed. Inlining is optional for a server, so a digest-only response yields `Some(empty)` → `is_hydrated()` returns true → the blob is never fetched → **a cache hit replays no task output**. Now keyed off whether the server actually sent bytes. This is independent of the above and affects any digest-only backend.

## Verification

Against Depot Cache, on a ~30-project repo:

- Before: `store failures: 6` of 6 tasks; nothing ever cached.
- After: `store failures: 0`, and with `.moon/cache` deleted entirely:

```
Tasks: 2 completed (2 cached)   Time: 731ms
Cache hit on manifest  storage="grpc-remote-cache"
Running cache hydration operation  hydrate_from=Storage(grpc-remote-cache)
```

Existing tests pass (`moon_cache_storage`, and `moon_task_runner` `output_hydrater_test`, which already covers the stdout-from-CAS path). Added two regression tests: one asserting the raw fields stay empty on upload, one asserting a digest-only response stays unhydrated so the blob is fetched.

Happy to adjust the approach for (2) — reading the hash manifest from `cache_dir` keeps the change small, but I am glad to plumb the bytes through from the task runner instead if you would prefer that.

### DIFF
--- a/crates/cache-storage/src/storage.rs
+++ b/crates/cache-storage/src/storage.rs
@@ -250,6 +250,7 @@ impl Storage {
                 digest.to_owned(),
                 manifest.clone(),
                 self.context.workspace_root.clone(),
+                self.context.cache_dir.clone(),
             ))));
         }
 
@@ -353,6 +354,7 @@ impl Storage {
                 digest.to_owned(),
                 manifest.clone(),
                 self.context.workspace_root.clone(),
+                self.context.cache_dir.clone(),
             ))));
         }
     }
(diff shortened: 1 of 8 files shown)


crates/cache-storage/src/storage.rs
```diff
@@ -417,8 +419,23 @@ async fn persist_manifest_in_backend(
...
 ) -> miette::Result<()> {
-    let blob_inputs = manifest.collect_blob_inputs(&workspace_root);
+    let mut blob_inputs = manifest.collect_blob_inputs(&workspace_root);
+
+    // The action digest is the content hash (and byte length) of the hash manifest that
+    // produced it, so that manifest *is* the blob the digest addresses. Upload it with
+    // the outputs: backends that validate the RE contract reject an action result whose
+    // action digest is absent from the CAS ("action digest <hash>/<size> not found in
+    // CAS"), because a client is expected to have uploaded it before referencing it.
+    let action_blob_path = cache_dir.join("hashes").join(format!("{}.json", *digest));
```
milesj: @KeyZer This feels like the wrong spot to do this. Should probably done in the task runner layer.

milesj: @KeyZer Think you could get around to this today?

milesj:
Awesome thanks!

PR #2624, opened 2026-07-16
feat: Store the workspace graph cache in remote storage backends

royreznik:
## Summary

Adds a new experiment, `experiments.remoteWorkspaceGraphCache` (`MOON_EXPERIMENT_REMOTE_WORKSPACE_GRAPH_CACHE`), that persists the serialized workspace graph to storage backends — most usefully a remote Bazel RE API backend — keyed by the existing graph fingerprint digest. On a local cache miss, the graph is hydrated from a backend instead of being rebuilt.

## Motivation

Since #2602, a workspace graph cache hit skips all `extend_project_graph` plugin calls. In our Go monorepo (~130 modules) that's the difference between ~0.2s and ~10s+ of startup for every moon command. But the graph state is local-only, so ephemeral CI runners rebuild it (and re-run `go list --deps` across the whole repo) on every single run — remote caching only covers task outputs. With this experiment, the first runner to build a given graph digest uploads ~one small JSON blob, and every other runner (and fresh clone/devcontainer) hydrates it.

## Implementation

- **Storage**: reuses the existing `Storage` manifest/blob APIs unchanged — the graph JSON is archived as a single-file `Manifest` keyed by the fingerprint digest (maps to `ActionCache` + CAS for the gRPC backend). Upload is awaited so graph-only commands don't drop it; failures are non-fatal and only log a warning.
- **Portability**: the graph is built before `sync_workspace` connects the backends, so the experiment path connects them itself (idempotent). `Project.root` is the only absolute path in the serialization closure (everything else is workspace-relative), so it's recomputed from `project.source` against the current workspace root whenever a cached graph is loaded — from remote *or* local. This also fixes stale roots when a workspace moves on disk, complementing the existing `in_docker` fingerprint flag.
- **Robustness**: a cache file that fails to deserialize is now discarded and rebuilt instead of hard-erroring (previously a corrupt `workspaceGraph.json` was fatal).
- Both the sync and async builders are covered.

## Testing

Extended the existing `cache_tests!` suite (runs against both builders):
- `hydrates_graph_from_storage_on_other_machine` — two sandboxes (different absolute paths) sharing a mocked remote backend; asserts the second build performs zero `extend_project_graph` calls (via the tc-tier1 marker), state hashes match, and project roots point into the *second* workspace.
- `rehydrates_stale_project_roots_from_local_cache` — poisons serialized roots, asserts they're recomputed on load.
- `rebuilds_if_cached_graph_is_corrupt` — garbage cache file rebuilds instead of erroring.

The test mocker gained a `remote_storage_dir` option that registers a `LocalStorage` CAS as a *remote* backend.

Docs, JSON schema, TS types, and CHANGELOG updated. Marked draft to get feedback on the approach first — happy to adjust (naming, whether root rehydration should be unconditional, whether the upload should stay fire-and-forget).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/workspace/src/workspace_cache.rs
+++ b/crates/workspace/src/workspace_cache.rs
@@ -2,13 +2,17 @@ use crate::projects_builder::{ProjectBuildData, ProjectBuildDataMap};
 use crate::workspace_builder::WorkspaceBuilderContext;
 use miette::IntoDiagnostic;
 use moon_cache::{ContentHash, cache_item};
+use moon_cache_storage::{Manifest, ManifestFile};
 use moon_common::path::WorkspaceRelativePathBuf;
 use moon_common::{Id, is_docker};
 use moon_env_var::GlobalEnvBag;
 use moon_hash::{Digest, fingerprint};
 use rustc_hash::FxHashMap;
+use starbase_utils::fs;
 use std::collections::{BTreeMap, BTreeSet};
+use std::path::Path;
 use std::sync::Arc;
+use tracing::{debug, warn};
 
 cache_item!(
     pub struct WorkspaceGraphCacheState {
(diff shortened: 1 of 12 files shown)


crates/workspace/src/workspace_cache.rs
```diff
@@ -209,3 +213,142 @@ pub async fn generate_graph_cache_digest(
...
+    if !context
+        .workspace_config
+        .experiments
+        .remote_workspace_graph_cache
+        || !context.cache_engine.storage.is_remote_enabled()
+    {
+        return false;
+    }
+
+    if let Err(error) = context.cache_engine.storage.connect_backends().await {
```
milesj, line 232: This shouldn't be done here, probably in the session analyze phase.

crates/workspace/src/workspace_cache.rs
```diff
@@ -209,3 +213,142 @@ pub async fn generate_graph_cache_digest(
...
+/// and non-fatal, as this is purely an optimization over rebuilding.
+pub async fn load_graph_from_storage(
+    context: &WorkspaceBuilderContext,
+    digest: &Digest,
+    cache_path: &Path,
+) -> bool {
+    let storage = &context.cache_engine.storage;
+
+    let result: miette::Result<bool> = async {
+        let Some(source) = storage.load_manifest(digest).await? else {
```
milesj, line 254: The manifest should absolutely not be used, as that represents a task execution state, and has nothing to do with the workspace graph.

There's a reason why storage is only used for a tasks, as it wasn't designed for non-task related cache.

We need an entirely different solution here.

milesj: Adding some more APIs here: https://github.com/moonrepo/moon/pull/2626/changes

royreznik: Thanks — agreed on both points. I'll move the backend connection into the session analyze phase, and drop the `Manifest` usage in favor of the new blob APIs from #2626.

Before I rewrite it, one design question on how the graph blob should be keyed, since the new `store_blob`/`retrieve_blob` map to pure CAS (`BatchUpdateBlobs`/`BatchReadBlobs`): the natural lookup key here is the workspace graph *fingerprint* digest (hash of the inputs), but a CAS blob must be keyed by the hash of its own *content* — RE API servers verify this on upload (and #2622 just tightened compliance), so storing the graph JSON under the fingerprint digest would be rejected by strict servers like bazel-remote. And with content addressing alone, a second machine can't find the blob, because it only knows the fingerprint.

Options I see:

1. **Store the blob under the fingerprint digest anyway.** Zero new APIs, works with the local CAS backend, but violates the RE contract and fails against compliant remotes — probably a non-starter.

2. **Add a small keyed-state concept to `StorageBackend`** — e.g. `store_keyed_blob(key: Digest, blob)` / `retrieve_keyed_blob(key)`. The gRPC backend implements it via the ActionCache (an `ActionResult` whose single output file references the content-addressed CAS blob), since AC is the RE API's only fingerprint→content mapping. Internally it stays its own concept — not the task `Manifest` — so task execution state and workspace state don't mix. Local backend just writes a keyed file. This is compliant and reusable for any future non-task state moon wants to cache remotely.

3. **Scope the experiment to backends that support non-content-addressed keys** (e.g. a future moonbase KV, or the HTTP backend if it relaxes verification), and skip it for gRPC remotes for now.

My preference is (2) — it keeps the RE contract intact and gives moon a clean primitive for non-task state — but you know where you want the storage layer to go better than I do. Happy to implement whichever direction you prefer.

crates/workspace/src/workspace_cache.rs
royreznik: Agreed — will move the connection into the session analyze phase (gated on the experiment + remote being configured), and keep this function a pure check.

crates/workspace/src/workspace_cache.rs
milesj: I figured this would be 2 CAS entries, the fingerprint, and then the graph cache itself. In the future, I want to move state fields to the CAS anyways.

Manifests are already the key-value store (the hashes don't match the blob bytes), we simply use the same CAS engine under the hood for simplicity, but it's not used like a CAS. Right now manifests are tied to tasks and changing that is a bigger lift.

royreznik: Sounds good — I'm happy to rework the PR along the 2-entry approach, and it's great that this lines up with moving state fields to the CAS later.

One implementation detail I want to get right before pushing code: the *link* entry (fingerprint → graph blob). The graph blob itself is content-addressed, so it's a clean CAS entry everywhere. But if the link entry goes through `store_blobs` → `BatchUpdateBlobs` keyed by the fingerprint digest, strict remotes will reject it — bazel-remote verifies that uploaded bytes hash to the declared digest, so any entry whose "hash doesn't match the blob bytes" only works via the ActionCache route (which is how manifests get away with it today) or against the local engine (which doesn't verify).

So, two ways to do your 2-entry scheme, and I'm glad to implement either:

1. **Link entry rides the AC**, like manifests do — a minimal backend method (e.g. `store_state_blob(key, bytes)` / `retrieve_state_blob(key)`) that the gRPC backend maps to `UpdateActionResult`/`GetActionResult` and the local backend maps to a keyed write. Not the task `Manifest` struct — just the same transport. Works against compliant remotes, and gives you the primitive for moving other state fields to the CAS later.

2. **Link entry as a plain CAS blob under the fingerprint digest** — smallest possible change on top of #2626, but it will fail against verifying remotes (bazel-remote and most RE-compliant servers), which is the main deployment target for this experiment.

If you'd rather take this over and shape the storage side yourself — totally fine by me, I can rebase the graph/builder half on whatever primitive you land. Otherwise, tell me which of the two you want and I'll push the rework.

milesj: Sorry for the late response to this, still trying to figure things out. I've tried to make the manifest layer generic and not tied to tasks, but it's been a bigger lift than expected. I thought that would be a nice path to go.

Manifests are the KV store. Blobs are the CAS store. But the bazel API contract makes it not possible.

royreznik:
closing it, not progressing

milesj:
Just an update here, but I made progress for this by storing hashes in local storage: https://github.com/moonrepo/moon/pull/2716

PR #2650, opened 2026-07-31
feat(pdk): define VCS provider API

lamalex:
## Overview

This PR adds the PDK contract for version control system (VCS) provider plugins.
This PR does not load or run VCS plugins.

## Changes
- Add provider registration and metadata types.
- Add workspace detection types.
- Add types for a fixed repository observation.
- Add exact state IDs and movable VCS references.
- Add current, recorded, and baseline states.
- Add repository and client metadata.
- Add working and submission impact requests.
- Add file change masks and result completeness states.
- Add optional hook environment types.
- Add protocol version checks for host and provider compatibility.
- Add serialization tests for the wire contract.

A provider must keep one fixed observation for each plugin instance. Later queries must use that observation. They must not read newer repository state.
## Purpose
This contract lets Moon support VCS providers without hard-coded provider logic.
Later PRs will add the plugin host, provider runtime, Git provider, and Jujutsu provider.

### DIFF
--- a/crates/pdk-api/src/vcs.rs
+++ b/crates/pdk-api/src/vcs.rs
@@ -15,7 +15,6 @@
 use crate::{Id, MoonContext, VirtualPath};
 use bitflags::bitflags;
 use std::collections::BTreeMap;
-use std::fmt;
 use std::path::PathBuf;
 use warpgate_api::{api_enum, api_struct, api_unit_enum};
 
@@ -27,98 +26,6 @@ use warpgate_api::{api_enum, api_struct, api_unit_enum};
 pub const VCS_PLUGIN_PROTOCOL_VERSION: u16 = 6;
 
 api_struct!(
-    /// Opaque provider-defined identity for an exact repository state.
-    #[serde(transparent)]
-    pub struct VcsStateId(pub String);
-);
-
-impl VcsStateId {
-    pub fn as_str(&self) -> &str {
-        &self.0
-    }
-
-    pub fn into_inner(self) -> String {
-        self.0
-    }
-}
-
-impl AsRef<str> for VcsStateId {
-    fn as_ref(&self) -> &str {
-        self.as_str()
-    }
-}
-
-impl fmt::Display for VcsStateId {
-    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        formatter.write_str(self.as_str())
-    }
-}
-
-impl From<String> for VcsStateId {
-    fn from(value: String) -> Self {
-        Self(value)
-    }
-}
-
-impl From<&str> for VcsStateId {
-    fn from(value: &str) -> Self {
-        Self(value.to_owned())
-    }
-}
-
-api_struct!(
-    /// Opaque provider expression that resolves to a repository state.
-    #[serde(transparent)]
-    pub struct VcsReference(pub String);
-);
-
-impl VcsReference {
-    pub fn as_str(&self) -> &str {
-        &self.0
-    }
-
-    pub fn into_inner(self) -> String {
-        self.0
-    }
-}
-
-impl AsRef<str> for VcsReference {
-    fn as_ref(&self) -> &str {
-        self.as_str()
-    }
-}
-
-impl fmt::Display for VcsReference {
-    fn fmt(&self, formatter: &mut fmt::Formatter<'_>) -> fmt::Result {
-        formatter.write_str(self.as_str())
-    }
-}
-
-impl From<String> for VcsReference {
-    fn from(value: String) -> Self {
-        Self(value)
-    }
-}
-
-impl From<&str> for VcsReference {
-    fn from(value: &str) -> Self {
-        Self(value.to_owned())
-    }
-}
-
-impl From<VcsStateId> for VcsReference {
-    fn from(value: VcsStateId) -> Self {
-        Self(value.into_inner())
-    }
-}
-
-impl From<&VcsStateId> for VcsReference {
-    fn from(value: &VcsStateId) -> Self {
-        Self(value.as_str().to_owned())
-    }
-}
-
-api_struct!(
     /// Input passed to `register_vcs` before any other provider operation.
     pub struct RegisterVcsInput {
         /// ID under which the host loaded this plugin instance.
@@ -127,16 +34,6 @@ api_struct!(
         pub host_protocol_version: u16,
     }
 );
-
-impl VcsState {
-    pub fn id_str(&self) -> Option<&str> {
-        self.id.as_ref().map(VcsStateId::as_str)
-    }
-
-    pub fn id_reference(&self) -> Option<VcsReference> {
-        self.id.as_ref().map(VcsReference::from)
-    }
-}
 
 api_struct!(
     /// Provider metadata returned by `register_vcs`.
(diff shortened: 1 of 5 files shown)


crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+///
+/// VCS plugins use lockstep protocol generations. Adding fields without safe
+/// serde defaults, changing lifecycle semantics, or assigning new change-mask
+/// bits requires incrementing this version.
+pub const VCS_PLUGIN_PROTOCOL_VERSION: u16 = 6;
+
+api_struct!(
+    /// Opaque provider-defined identity for an exact repository state.
+    #[serde(transparent)]
+    pub struct VcsStateId(pub String);
```
milesj: Maybe just `type StateId = String` since it doesn't provide any new functionality?

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+impl From<&str> for VcsStateId {
+    fn from(value: &str) -> Self {
+        Self(value.to_owned())
+    }
+}
+
+api_struct!(
+    /// Opaque provider expression that resolves to a repository state.
+    #[serde(transparent)]
+    pub struct VcsReference(pub String);
```
milesj: Same here.

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+
+    pub fn id_reference(&self) -> Option<VcsReference> {
+        self.id.as_ref().map(VcsReference::from)
+    }
+}
+
+api_struct!(
+    /// Provider metadata returned by `register_vcs`.
+    #[serde(default)]
+    pub struct VcsPluginMetadata {
```
milesj: `RegisterVcsOutput`

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+);
+
+api_struct!(
+    /// Applicability reported by `detect_vcs`.
+    ///
+    /// `active` means that this provider applies to the workspace and can
+    /// establish an observation. User configuration controls whether a plugin
+    /// is selected at all; there is no second enabled state after observation.
+    #[serde(default)]
+    pub struct DetectVcsOutput {
```
milesj: It needs to provide the working/worktree dir, and repository root paths, as `VirtualPath`s.

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+        #[default]
+        Unknown,
+    }
+);
+
+api_struct!(
+    /// Provider metadata and exact states pinned by `observe_vcs`.
+    pub struct VcsObservation {
+        /// Stable provider kind, such as `git` or `jj`.
+        pub provider: Id,
```
milesj: This is called `client` on the moon side.

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+        /// Version of the source-control client used by the provider.
+        pub client_version: Option<String>,
+        /// Absolute or virtual path to the repository root.
+        pub repository_root: String,
```
milesj: All path fields need to be `VirtualPath`.

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+);
+
+api_struct!(
+    /// Input passed to `setup_vcs_hook_environment` after observation.
+    pub struct SetupVcsHookEnvironmentInput {
+        /// Must be identical to the observation context.
+        pub context: MoonContext,
+        /// Canonical path to Moon's hooks directory, relative to the workspace.
+        pub hooks_dir: String,
+        /// Provider-native hook names that Moon intends to install.
```
milesj: moon is always lowercase

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+api_struct!(
+    /// Output returned by `get_vcs_impacts`.
+    #[serde(default)]
+    pub struct GetVcsImpactsOutput {
+        /// Canonical UTF-8 file paths relative to the Moon workspace root.
+        ///
+        /// Keys use `/` separators and must be non-empty, must not contain `.`,
+        /// `..`, empty components, backslashes, NULs, or names invalid on the
+        /// host platform, and must not be absolute.
+        pub changes: BTreeMap<String, VcsChangeMask>,
```
milesj: Use `PathBuf` for the key.

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+api_struct!(
+    /// Input passed to `get_vcs_impacts` after observation.
+    pub struct GetVcsImpactsInput {
+        /// Must be identical to the context used to establish the observation.
+        pub context: MoonContext,
+        pub intent: VcsImpactIntent,
+    }
+);
+
+api_struct!(
```
milesj: Probably just use https://crates.io/crates/bitflags

crates/pdk-api/src/vcs.rs
```diff
@@ -0,0 +1,522 @@
...
+        Incomplete,
+        /// The provider cannot determine whether history is complete.
+        #[default]
+        Unknown,
+    }
+);
+
+api_struct!(
+    /// Provider metadata and exact states pinned by `observe_vcs`.
+    pub struct VcsObservation {
```
milesj: Still a bit confused when the observation triggers.

crates/pdk-api/src/vcs.rs
lamalex: Not a fan of the [Newtype pattern](https://doc.rust-lang.org/rust-by-example/generics/new_types.html)? I used transparent newtypes because there are a set of semantic Ids (ex: `VcsStateId` and `VcsReference`). They use strings on the wire, but they should not be interchangeable in Rust. The newtypes catch that mistake at compile time and add no runtime or serialization overhead.

If you think that distinction does not justify the added API surface, I can simplify them to String. Is your main concern ergonomics, or the value of the distinction itself?

milesj: > but they should not be interchangeable in Rust

This won't happen on the moon side, and developers would simply use `as_str` + `new` as a means to work around it, so it really doesn't save us anything.

crates/pdk-api/src/vcs.rs
lamalex: I renamed `observe_vcs` to `initialize_vcs` and `VcsObservation` to `InitializeVcsOutput`.

I also documented the lifecycle: moon calls `initialize_vcs` once when it creates the VCS client for a command, after detection and before any impact or hook queries. The initialized state remains fixed for that plugin instance. A new state requires a new plugin instance.

milesj: @lamalex The only reason I ask that it seems like all 3 of these functions basically run at the same time.

Load the plugin (`register_vcs`) -> detect if the VCS is active (happens right after loading basically) (`detect_vcs`) -> then capture the observation (right after detecting?) (`initialize_vcs`).

May make sense to merge detect/init? Or defer the init until the first "get changed files" occurs?

lamalex: > May make sense to merge detect/init?

Yeah, this is a good call. Separate detect/init is a wart from the initial PR in which the `jj` plugin tried to be an extension of the built-in git behavior vs. turning the entire Vcs support into something pluggable.

milesj:
@lamalex looks like you need to rebase

milesj:
@lamalex Changes look good, just need to rebase.

I updated to proto v0.60 which changed some PDK APIs. Just ensure everything still compiles.

PR #2654, opened 2026-08-05
fix: prevent cache creation on completions

JuanoD:
.moon/cache was being created when generating completions

### DIFF
--- a/crates/app/src/session.rs
+++ b/crates/app/src/session.rs
@@ -502,10 +502,9 @@ impl AppSession for MoonSession {
         }
 
         // Ensure all in-flight storage tasks have finished
-        self.get_cache_engine()?
-            .storage
-            .wait_for_background_tasks()
-            .await?;
+        if let Some(engine) = self.cache_engine.get() {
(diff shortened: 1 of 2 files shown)


crates/app/src/session.rs
```diff
@@ -502,10 +502,12 @@ impl AppSession for MoonSession {
         }
 
         // Ensure all in-flight storage tasks have finished
-        self.get_cache_engine()?
-            .storage
-            .wait_for_background_tasks()
-            .await?;
+        if let Some(engine) = self.cache_engine.get() {
```
milesj: @JuanoD A better option is to maybe wrap with `if self.requires_workspace_configured() {}`

JuanoD: The reasoning was to let cache_engine finish its work if it actually existed, instead of spawning one when not needed.

JuanoD: @milesj you let me know if that reasoning is correct or there is a different change needed

milesj: Sure makes sense.

Issue #2662, opened 2026-08-13
[bug] docker prune seems to ignore toolchain failures


nhomble:
**Describe the bug**

`moon docker prune` swallows errors from the toolchain plugins it runs

**Steps to reproduce**

https://github.com/nhomble/moon-2662

**Expected behavior**

if the toolchain fails (like `pnpm`) as part of prune I think the error should propagate (so that my CI step that runs the image build fails)

**Screenshots**

**Environment**

```

```

**Additional context**

nhomble:
https://github.com/moonrepo/moon/pull/2663#issuecomment-5283128930

PR #2663, opened 2026-08-13
fix: prune should propagate toolchain errors

nhomble:
https://github.com/moonrepo/moon/issues/2662


milesj:
This is actually fixed in the v2.5 branch, which will be landing soon.

https://github.com/moonrepo/moon/pull/2637

Appreciated though!

nhomble:
My bad Miles! I had tunnel vision on 2.4.6 while triaging our issues 😓 

I'm glad you're already ahead of this

Issue #2691, opened 2026-08-26
`moon docker setup` fails locating Node.js executable when running Python's `SetupEnvironment` in polyglot workspaces


SavioR0:
**Describe the bug**

In a polyglot workspace where `.moon/toolchains.yml` declares **both** a Python/uv toolchain and a Node/JavaScript toolchain, running `moon docker setup` for a Python-only focused project (`language: python`) fails during the `SetupEnvironment(unstable_python, <root>)` action with:

```
Error: proto::locate::missing_executable

  × Unable to find an executable for Node.js, expected file ~/.proto/tools/
  │ node/22.23.2/bin/node does not exist.
```

The action graph correctly schedules `SetupToolchain(unstable_python)` and `SetupToolchain(unstable_uv)` for the Docker build — it does **not** schedule `SetupToolchain(node)` or `SetupToolchain(pnpm)`, since the focused project has no dependency on those toolchains. However, `SetupEnvironment` for Python still attempts to **locate** the executables of every toolchain declared in the workspace, including Node/pnpm, and fails hard when they aren't present on disk (because their own `SetupToolchain` action never ran).

This was found while verifying the fix for #2381 on 2.5.2 — the original dependency-installation bug is indeed fixed, but this new issue blocks Docker builds entirely in any repo that mixes a Python backend with a Node/JS frontend in the same moon workspace, which is a fairly common monorepo shape.

**Steps to reproduce**

I reproduced this using the minimal repro repository originally created by @lucasmsoares96 for #2381 (a Python workspace with `python_bot` depending on `chatwoot_api`).

1. `.moon/toolchains.yml` declares both toolchains:
   ```yaml
   unstable_python:
     version: 3.13.0
     packageManager: 'uv'
   unstable_uv:
     version: '0.9'

   node:
     version: '22'
   javascript:
     packageManager: pnpm
   pnpm:
     version: 10.28.1
   ```
   Note: no Node/JS project actually exists anywhere in the workspace — the toolchains are simply declared.

2. `python_bot/moon.yml`:
   ```yaml
   stack: backend
   language: python
   layer: application
   id: python_bot-backend
   dependsOn:
     - chatwoot_api
   ```

3. Dockerfile (following the official [Docker integration guide](https://moonrepo.dev/docs/guides/docker) verbatim, no custom install scripts):
   ```dockerfile
   FROM node:latest AS base
   WORKDIR /app
[...]

PR #2693, opened 2026-09-02
fix: scope plugin commands to source toolchain

xiang-baseten:
might related to https://github.com/moonrepo/moon/issues/2691

## Summary

- preserve the originating toolchain when executing plugin-generated commands
- scope command hooks and executable-path augmentation to that toolchain
- add regression coverage proving unrelated toolchains are not invoked

## Root cause

Dependency install commands may run at a shared dependencies workspace root, so they do not always have a project or task context. Since Moon 2.5, `get_toolchain_ids(None, None)` returns every configured workspace toolchain for genuinely global commands.

The plugin command executor discarded the ID of the toolchain that generated a command and then inferred toolchains only from its project/task context. For a shared Go workspace command such as `go work sync`, both values were `None`, so Moon augmented it with Go, Node, Python, Rust, and every other configured toolchain. In a minimal Docker layer where only Go had been installed, resolving the unrelated Python command path failed before the Go command could run.

## Fix

Add the originating toolchain ID to `ExecCommandOptions` and pass it through the serial, parallel, cache, and retry execution paths. Plugin-generated commands use this explicit ID when applying toolchain hooks and executable paths. Commands without an explicit source retain the existing project/task/global inference behavior.

This preserves the all-toolchains behavior required by genuinely global commands while preventing a command generated by one toolchain from depending on unrelated toolchains. It also leaves toolchain setup parallelism unchanged.

## Validation

I replicate the issue locally and use patch moon which build successfully.

### DIFF
--- a/crates/actions/src/actions/install_dependencies.rs
+++ b/crates/actions/src/actions/install_dependencies.rs
@@ -184,6 +184,7 @@ pub async fn install_dependencies(
     let options = ExecCommandOptions {
         project,
         prefix: action.get_prefix().into(),
+        toolchain_id: Some(toolchain.id.clone()),
         working_dir: Some(deps_root),
         on_exec: Some(Arc::new(move |cmd, attempts| {
             handle_on_exec(&console, cmd, attempts)
--- a/crates/actions/src/actions/setup_environment.rs
+++ b/crates/actions/src/actions/setup_environment.rs
@@ -131,6 +131,7 @@ pub async fn setup_environment(
             ExecCommandOptions {
                 project,
                 prefix: action.get_prefix().into(),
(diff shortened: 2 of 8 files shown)


xiang-baseten:
@milesj might not be the best fix as I am not familiar with codebase, let me WDYT.

milesj:
@xiang-baseten This is close but not fully correct I believe. The reason is that toolchains can depend on other toolchains (npm requires node, etc), so this would filter out the required ones.

I'm looking into this right now. Hopefully it's a simple fix.

milesj:
Ok I think I fixed this in v2.5.4.

PR #2697, opened 2026-09-07
fix: Pack and restore symlinks in directory outputs without dropping node_modules

jbmusso:
## Problem

With the `casOutputsCache` experiment enabled, a task whose declared output directory contains symlinks completes successfully and is then failed by the archiver:

```
hash::read_stream
  × Failed to hash bytes from readable stream.
  ╰─▶ Is a directory (os error 21)
```

`ManifestPacker::insert_dir` walks the output with a files-only glob whose filters pass symlinks through, and `insert_file` then follows the link — a directory target (e.g. any pnpm `node_modules` layout, such as a Next.js standalone build) streams a directory into the hasher. Three adjacent defects sit on the same path:

- The glob walk applies the global negations, so `node_modules/**` inside a directory output is **silently dropped from the manifest** — a cache hit restores the output without its dependency store, exit 0, no warning.
- Relative symlink targets are rejected with `manifest::symlink_outside_workspace`.
- `ManifestUnpacker` restores links with absolute targets, which dangle when the output is copied out of the workspace (e.g. `COPY` into a container image).

The remote REAPI path shares this packer, so it is affected identically.

**To see the failures, check out the first commit** (`test: Reproduce directory-output packing failures for symlinks and node_modules`): `cargo test -p moon_manifest --no-fail-fast` fails 3 packer tests (EISDIR, wrong error for a relative target, missing `node_modules` entries) and 2 unpacker tests (relative link restoration).

## Solution

- `insert_dir`: enumerate entries with `fs::read_dir_all` instead of a glob (no implicit global negations) and classify each entry, routing symlinks to `insert_symlink`.
- `insert_symlink`: resolve relative targets against the link's parent before the workspace-containment check, and record the target workspace-relative.
- `ManifestUnpacker`: recreate links with targets relative to the link's location, so restored trees are position-independent.

`cargo test -p moon_manifest` green, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

*Disclaimer: this analysis and patch were AI-assisted.*


milesj:
@jbmusso Do you have an example of when you are caching `node_modules`? We suggest not doing that.

PR #2699, opened 2026-09-09
fix: preserve outputs when cache hydration fails

### LINKED ISSUE #2698: [bug] A failed remote cache read fails a task that ran and passed, and deletes its outputs
**Describe the bug**

When a remote cache read fails part way through, moon can mark a task as failed although the task ran and passed. The task prints a pass checkpoint, the stats line counts a fail, and the process exits 1.

The cause is the order of two steps.

`OutputHydrater::hydrate` calls `delete_existing_outputs()` **before** it downloads the blobs. If the download then fails, moon discards the restore and runs the task again, which is correct. But a task does not always recreate every declared output, because the output was already on disk before the run. `OutputArchiver::archive` then returns `TaskRunnerError::MissingOutputs`.

The action status is then written twice, from two places:

* `crates/actions/src/actions/run_task.rs` sets `action.status = result.operations.get_final_status()`. The last process operation is the task execution, which passed.
* The same function also returns `Err(error)`, and `crates/action-pipeline/src/action_runner.rs` replaces the status with `ActionStatus::Failed`.

So one action reports a pass in the checkpoint and a fail in the counter.

`ciReport.json` from the reproduction below shows both halves at once:

```json
{
  "label": "RunTask(pkg:bundle)",
  "status": "failed",
  "error": "Task pkg:bundle defines outputs but after being ran, either none or not all of them exist.",
  "operations": [
    { "meta": "hash-generation",  "status": "passed"  },
    { "meta": "output-hydration", "status": "invalid" },
    { "meta": "task-execution",   "status": "passed"  }
  ]
}
```

Reproduced on **2.4.5** and **2.5.4**.

**Steps to reproduce**

The reproduction needs a remote cache that answers CAS reads with 502. This script is enough.

`cache_server.py`:

```python
import os, sys
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer

HERE = os.path.dirname(os.path.abspath(__file__))
ROOT = os.path.join(HERE, "store")
BREAK = os.path.join(HERE, "BREAK")
os.makedirs(ROOT, exist_ok=True)

def blob_path(path):
    parts = [p for p in path.split("/") if p]
    return os.path.join(ROOT, parts[1] + "_" + parts[2]) if len(parts) == 3 else None

class Handler(BaseHTTPRequestHandler):
    protocol_version = "HTTP/1.1"
    def log_message(self, fmt, *a): pass

    def _empty(self, code):
        self.send_response(code)
        self.send_header("Content-Length", "0")
        self.end_headers()

    def do_GET(self):
        if "/cas/" in self.path and os.path.exists(BREAK):
            return self._empty(502)              # the cache dies mid restore
        p = blob_path(self.path)
        if p and os.path.exists(p):
            body = open(p, "rb").read()
            self.send_response(200)
            self.send_header("Content-Length", str(len(body)))
            self.end_headers()
            self.wfile.write(body)
        else:
            self._empty(404)

    def do_PUT(self):
        body = self.rfile.read(int(self.headers.get("Content-Length") or 0))
        p = blob_path(self.path)
        if p: open(p, "wb").write(body)
        self._empty(200)

    def do_HEAD(self):
        p = blob_path(self.path)
        self._empty(200 if (p and os.path.exists(p)) else 404)

ThreadingHTTPServer(("127.0.0.1", 8099), Handler).serve_forever()
```

`.moon/workspace.yml`:

```yaml
projects:
  pkg: 'pkg'
remote:
  api: http
  host: http://127.0.0.1:8099
  cache:
    verifyIntegrity: true
experiments:
  casOutputsCache: true
```

`pkg/moon.yml`:

```yaml
layer: library
language: bash
tasks:
  build:
    command: 'bash ./build.sh'
    inputs: ['src/**/*', 'build.sh']
    outputs: ['out']
  bundle:
    command: 'bash ./bundle.sh'
    deps: ['build']
    inputs: ['src/**/*', 'bundle.sh']
    outputs: ['extra']
```

`pkg/build.sh` writes some files into `out/`. `pkg/bundle.sh` passes but does not touch `extra/`, because `extra/` is produced elsewhere and this task only caches it:

```bash
#!/usr/bin/env bash
set -e
echo "BUNDLE RAN OK"
```

Then:

1. `mkdir -p pkg/extra && echo artifact > pkg/extra/thing.txt`
2. Start the server. Make sure that the file `BREAK` does not exist.
3. Make an uncommitted edit under `pkg/src/` so the task is affected.
4. `moon ci pkg:bundle`. This passes and fills the cache.
5. `touch BREAK` and `rm -rf .moon/cache`.
6. `moon ci pkg:bundle`.

Step 6 gives:

```
[ WARN ] moon_cache_storage::storage_backend  Failed to retrieve blobs, ...
         expected_count=2 actual_count=0
         errors=["Failed to make HTTP request (502 Bad Gateway)."]
▮▮▮▮ pkg:bundle (85e91fb2)
pkg:bundle | BUNDLE RAN OK
▮▮▮▮ pkg:bundle (13ms, 85e91fb2)

Tasks: 1 completed, 1 failed
```

Exit code 1. Both tasks print a pass checkpoint. On 2.4.5 no section names the failed task and no error text appears anywhere in the output.

**Expected behavior**

A remote cache read that fails is a cache miss. Moon already treats it as one and runs the task again, which is right. After that the run must not fail, and the outputs that were on disk before the run must still be there.

Two smaller points in the same area:

1. `delete_existing_outputs()` runs before the download. A failed download therefore removes outputs that were on disk and cannot put them back. A delete after a complete download avoids this.
2. On 2.5.4 the `REVIEW` section prints the target and nothing else. `action.error` holds the reason and reaches `ciReport.json`, but no one prints it.

**Environment**

```
  System:
    OS: Linux 6.8 Ubuntu 22.04.5 LTS
    CPU: (4) arm64 Neoverse-V2
  Binaries:
    Node: 24.19.0
    npm: 11.17.0
    pnpm: 11.24.0
  moon: 2.4.5 and 2.5.4
```

**Additional context**

Two related findings from the same reproduction.

**Glob outputs lose data without any error.** Change `bundle.outputs` to `['extra/**']` and repeat the steps. The run exits 0 and reports success, but the output directory is empty afterwards, and moon uploads a manifest for that empty result. Same on both versions:

```
before broken run: pkg/extra has 1 file(s)
broken run exit=0     Actions: 4 completed
after  broken run: pkg/extra has 0 file(s)
```

This one worries me more than the first, because a poisoned cache entry then spreads to every other machine.

**`moon ci` drops its own summary on 2.4.5.** `crates/app/src/commands/ci.rs` sets `exec.summary = Some(Some(SummaryOption::Detailed))`, but `crates/app/src/helpers.rs` reads the summary back from `session.cli.command`, which still holds the unmodified `CiArgs`. The pipeline therefore receives `summary: None` and prints neither `REVIEW` nor `SUMMARY`. 2.5.4 passes the value as a parameter, so this part is already fixed. I mention it because it is why the first bug was invisible for us: a red build, a green log, and no named task.

What led us here: our self-hosted `bazel-remote` ran out of memory during a CI run and answered 502 for about two minutes. Every task in two CI jobs passed, and both jobs still exited 1 with `1 failed` and nothing to point at.

Tyagiquamar:
## What does this change do?

Preserves existing task outputs when cache hydration cannot retrieve a required blob.

## Why is this change needed?

Hydration deleted outputs before all cache reads completed. If a remote or local cache blob was missing, the task could be left without outputs it was expected to recreate, causing a later `MissingOutputs` failure. Outputs are now deleted only after the cache manifest and blobs have been retrieved successfully.

Fixes moonrepo/moon#2698.

## How has this been tested?

- `cargo test -p moon_task_runner --test output_hydrater_test preserves_existing_output_when_cache_blob_is_missing` in Docker with Rust 1.98.1
- Result: 1 passed, 0 failed
- `cargo fmt --check --all`
- `git diff --check`


milesj:
Thanks!

PR #2700, opened 2026-09-11
feat(plugin): Add the native process host required by VCS plugins

lamalex:
Next PR on the VCS plugin train:

This adds the host-side APIs that VCS plugins use to execute native commands through a limited, capability-checked process interface. Plugins declare required executables during registration, and the host validates and resolves those declarations before allowing execution.

The implementation uses std::process::Command::output() and does not provide:

- Process runtime limits
- Total capture limits
- Output-drain deadlines
- Process-tree termination
- Starbase registry cancellation or shutdown integration

These lifecycle guarantees remain deferred to Starbase / starbase_process. That work is progressing in parallel, but does not block development and controlled use of Moon’s bundled VCS providers.


crates/pdk-api/src/process.rs
```diff
@@ -0,0 +1,132 @@
...
+        /// Plugin-local identifier used by subsequent process requests.
+        pub id: Id,
+        /// Executable name to resolve through the host's PATH.
+        pub executable: String,
+    }
+);
+
+api_struct!(
+    /// An opaque process command requested by an opt-in plugin.
+    pub struct ProcessCommandInput {
```
milesj, line 18: Can you explain what the difference between this and the existing https://github.com/moonrepo/proto/blob/master/crates/warpgate-api/src/host_funcs.rs#L75 would be?

There's a lot of overlap here.

lamalex: I implemented it this way because native subprocesses execute outside the WASM sandbox. The API I added lets the host enforce a fixed executable, direct argv without a shell, and a workspace-confined cwd. I was trying to limit the impact of a confused-deputy bug where untrusted repository data influences process execution.

Examples:
1. A provider might derive a working directory from repository worktree or submodule metadata. A crafted repository could make that path resolve through a symlink outside the workspace, causing a native command with filesystem side effects to run there. Warpgate would accept that cwd, while this API canonicalizes it and rejects the command before spawning.
2. repository-controlled input cannot select `sh` or\` as the executable or be evaluated as shell source.

But, using Warpgate is also a valid approach so long as plugins are first-party, trusted, and checksum-verified.

The trade-offs as I see them:

Security: The new API removes easy command/path/shell injection and enforces cwd confinement. It does not make arbitrary third-party plugins safe: capabilities are currently self-declared, environment variables remain flexible, and Git/JJ can launch other processes themselves. Warpgate has a broader attack surface, but maybe that's fine?

Complexity: Reusing Warpgate would remove a meaningful amount of custom host API and test code. Keeping this interface duplicates some process execution logic, but establishes the constrained contract we would need before opening VCS providers to third parties. (blocked on the `starbase_process` work that actually prevents a lot of the bad stuff host processes could do).

moon is your project, so whatever you want I'm happy to follow, but that's at least the rationale for the new API and why Warpgate wasn't re-used

crates/plugin/src/host.rs
```diff
@@ -87,6 +318,225 @@ pub fn create_host_functions(data: MoonHostData, shared_data: HostData) -> Vec<F
...
+        .map_err(|error| {
+            Error::msg(format!(
+                "failed to execute process capability `{}` with `{}`: {error}",
+                input.capability,
+                executable.display()
+            ))
+        })
+}
+
+fn read_process_output_v1(
```
milesj, line 475: Is this just re-reading the output of a previously executed command? When will this be used?

We already have the process cache, it should maybe just piggyback off that.

crates/plugin/src/host.rs
```diff
@@ -87,6 +318,225 @@ pub fn create_host_functions(data: MoonHostData, shared_data: HostData) -> Vec<F
...
+
+        if !candidate.starts_with(workspace_root) {
+            return Some(candidate);
+        }
+    }
+
+    None
+}
+
+fn is_executable_file(path: &Path) -> bool {
```
milesj, line 377: Theres utils in starbase for this.

crates/plugin/src/host.rs
```diff
@@ -87,6 +318,225 @@ pub fn create_host_functions(data: MoonHostData, shared_data: HostData) -> Vec<F
...
+
+    is_single_component
+        && (!cfg!(windows) || !(lowercase.ends_with(".bat") || lowercase.ends_with(".cmd")))
+}
+
+fn resolve_process_executable(executable: &str, workspace_root: &Path) -> Option<PathBuf> {
+    resolve_process_executable_in_path(executable, workspace_root, &std::env::var_os("PATH")?)
+}
+
+fn resolve_process_executable_in_path(
```
milesj, line 345: `system_env::find_command_on_path`

crates/plugin/src/host.rs
```diff
@@ -87,6 +318,225 @@ pub fn create_host_functions(data: MoonHostData, shared_data: HostData) -> Vec<F
...
+
+fn validate_process_cwd(cwd: &Path, workspace_root: &Path) -> Result<(), Error> {
+    if !cwd.is_dir() {
+        return Err(Error::msg(format!(
+            "plugin process working directory `{}` is not a directory",
+            cwd.display()
+        )));
+    }
+
+    if !cwd.starts_with(workspace_root) {
```
milesj, line 531: I'm dubious on this. If the moon workspace is nested within a repo, then the moon workspace root != the repo root.

Are we 100% certain we wouldn't need to run commands at the repo root.

crates/pdk/src/vcs.rs
```diff
@@ -0,0 +1,88 @@
...
+    result: &ProcessCommandResult,
+    stream: ProcessOutputStream,
+    expected_len: u64,
+) -> AnyResult<Vec<u8>> {
+    let mut offset = 0u64;
+    let mut output = Vec::new();
+
+    loop {
+        let chunk = unsafe {
+            read_process_output_v1(Json(ProcessOutputChunkInput {
```
milesj, line 50: I'm curious what the performance of this would be. 

Looping through and constantly going across the guest/host bridge, or just calling once and pulling everything into memory.

crates/pdk-api/src/process.rs
milesj: Some of these, like the dir/exe validation can be bubbled up to warpgate. Would there be any other capabilities?

As whether to move forward with this or not, I left some comments in other parts of the code for more clarification.

Issue #2701, opened 2026-09-11
[feature] Improve performance of task hashing by using memoization for changed-file map


rupesh-one:
> AI disclosure. I used AI assistance to find this regression, to build the reproduction, and to write this report. I ran the reproduction myself and checked every code reference below.

# Task hashing rebuilds the same changed-file map once per task

## Summary

Moon caches the output of `git status`, but not the parse of it. Every task hash calls `get_changed_files`, which re-tokenises the cached output, rebuilds a`ChangedFiles` map, merges it, and converts every path. The answer cannot change within a run, so all of that work after the first call is repeated.

The cost scales with the number of changed files times the number of tasks. It is invisible in a profile, because the merge and the path conversion carry no spans.

## Evidence

Moon 2.5.2, linux/arm64, 4 CPU. A Docker build of one service in a large pnpm monorepo. 175 tasks in the run, all remote cache hits, so no build command executes.

The `git status` that every task re-parses:

```
bytes=31141105        # 31.1 MB
entries=279856
index_files=49626      # tracked files in the index
```

The subprocess itself runs once and takes 4.2s. Spans from `moon run --dump`:

| Span | self | total | n (polls) |
| --- | --- | --- | --- |
| `task_runner::hash` | 1184.0s | 1184.3s | 2430 |
| `git::tree::exec_status` | 192.2s | 192.2s | 4948 |
| `sha256::from_file` | 0.2s | 1.0s | 11436 |
| `git::tree::exec_ls_files` | 0.2s | 0.2s | 929 |

File content hashing is 1.0s in total, so it is not the cost. `exec_status` is 192s, and it covers only the tokenise-and-map step. The merge and the path conversion that follow it have no spans, and they account for a large part of the ~890s of `hash` that nothing else explains.

## Cause

`create_command` in `crates/vcs/src/git/common.rs` caches the subprocess, with the intent stated in a comment:

```rust
// Always cache the output of git commands, as they are expensive to run
// and we often run the same commands multiple times (e.g. `git status`).
command.set_cache(true);
```

`exec_capture_output` honours that and returns a cloned `Output` on a hit. The parse, however, happens after the await in `GitTree::exec_status`, on every call: `output_to_string` allocates a fresh `String` of the whole output, the tokens are matched against `STATUS_PATTERN`, and a `PathBuf` is joined and
inserted per record.

`GitClient::get_changed_files` then adds no memoisation of its own:

```rust
for tree in self.get_all_trees() {
[...]

PR #2702, opened 2026-09-11
perf: collect working tree status once per run

rupesh-one:
Fixes the issue described in https://github.com/moonrepo/moon/issues/2701.

## Problem

`GitClient::get_changed_files` spawned a `git status` per tree, merged the results and converted every path, on every call. The command output is already cached, so the subprocess ran once, but the parse, the merge and the path conversion repeated for each caller.

Task hashing calls this once per task, and the working tree cannot change while a run is in progress, so every call after the first recomputes an answer that is already known.

In one Docker build of a large monorepo, `git status` returned 279,856 entries over 31 MB. Across a 175-task run that rebuilt roughly 49 million map entries.

## Change

Hold the result in a `tokio::sync::OnceCell` on the client. The first caller collects the status; the rest clone it.

`ChangedFiles` gains `Clone` so callers still receive an owned value. The public signature of `get_changed_files` is unchanged.

## Correctness

The cache lives on the client instance, so its lifetime matches the run rather than the process.

This relies on the working tree not changing under a run, which is the same assumption the existing command-output cache already makes: that cache is enabled in `create_command` with the comment "Always cache the output of git commands, as they are expensive to run and we often run the same commands multiple times (e.g. `git status`)". This change extends that assumption from the raw bytes to the parsed value.

If that assumption is ever unsafe, the existing output cache is unsafe too, and both should be revisited together.

## Verification

`cargo check -p moon_vcs` passes.

Measuring the improvement end to end needs a build where the changed-file set is large. The companion fix that scaffolds `.gitignore` shrinks that set, so the two are best measured separately before they are combined.


milesj:
@rupesh-one Tests are failing.

rupesh-one:
> @rupesh-one Tests are failing.

Will work on fixing the tests today

milesj:
Awesome thanks!

PR #2707, opened 2026-09-14
fix(query): respect CI environment when tracking affected tasks and projects

### LINKED ISSUE #2706: [bug] `moon query tasks --affected` ignores CI when it checks `runInCI`
**Describe the bug**

In CI, `moon query tasks --affected` and `moon run --affected` disagree on which tasks are affected. The query checks `runInCI` as if it ran locally. So it drops `runInCI: only` tasks, which `moon run` does run in CI, and keeps `runInCI: false` tasks, which `moon run` skips.

`moon query affected` behaves the same way.

We use the query to build a GitHub Actions matrix before any worker starts, then each job runs `moon run <target> --affected`. All our publish tasks are `runInCI: only`, so the matrix never sees them as affected.

The affected tracker starts with `ci: false`:

https://github.com/moonrepo/moon/blob/9c9498248132f7a72156eb8dc96efb19c44a1f2a/crates/affected/src/affected_tracker.rs#L49

and `is_task_affected` drops an `only` task when that flag is false, before it looks at any changed file:

https://github.com/moonrepo/moon/blob/9c9498248132f7a72156eb8dc96efb19c44a1f2a/crates/affected/src/affected_tracker.rs#L472-L483

`moon exec` sets the flag from the CI environment:

https://github.com/moonrepo/moon/blob/9c9498248132f7a72156eb8dc96efb19c44a1f2a/crates/app/src/commands/exec.rs#L190-L194

https://github.com/moonrepo/moon/blob/9c9498248132f7a72156eb8dc96efb19c44a1f2a/crates/app/src/commands/exec.rs#L469

The query commands never call `set_ci_check`:

https://github.com/moonrepo/moon/blob/9c9498248132f7a72156eb8dc96efb19c44a1f2a/crates/app/src/commands/query/tasks.rs#L91-L98

**Steps to reproduce**

```yaml
# .moon/workspace.yml
projects:
  app: app
```

```yaml
# app/moon.yml
tasks:
  default:
    command: echo default
    inputs: ["*.txt"]
  only:
    command: echo only
    inputs: ["*.txt"]
    options:
      runInCI: only
  never:
    command: echo never
    inputs: ["*.txt"]
    options:
      runInCI: false
```

```sh
echo a > app/file.txt
git init -b main && git add -A && git commit -m init
echo b > app/file.txt && git commit -am change
export CI=true MOON_BASE=HEAD~1
```

All three tasks have the same input, and that input changed.

```
$ moon query tasks --affected | jq -c '.tasks.app | keys'
["default","never"]

$ moon query affected | jq -c '.tasks | keys'
["app:default","app:never"]

$ moon run app:only --affected
▮▮▮▮ app:only (b7c3b9d4)
only
▮▮▮▮ app:only (8ms, b7c3b9d4)

$ moon run app:never --affected
│ CAUTION
│ No tasks affected by changed files. Unable to execute action pipeline.
│ For targets app:never.
```

The query lists `never` and leaves out `only`. `moon run` does the opposite.

**Expected behavior**

In CI, the query commands return the same affected tasks as `moon run --affected`. Calling `set_ci_check(is_ci())` where the query commands build their tracker would do it.

**Environment**

```
moon: 2.5.4 (also present on master)

System:
  OS: macOS 26.6.1
  CPU: (8) arm64 Apple M3
```

**Additional context**

@milesj in #2521 you said `runInCI` applies to every command in v2, so I read this as a bug rather than intended behaviour.

#2630 asks for a pre-execution view of the resolved `moon ci` graph. That would cover more of our use case very cleanly, but this bug stands on its own: the existing query commands already check `runInCI`, just with the wrong environment.

Tyagiquamar:
## What does this change do?

Ensures `moon query tasks --affected`, `moon query affected`, and `moon query projects --affected` check the CI environment when evaluating `runInCI` task options, aligning query results with `moon run --affected` and `moon exec` behavior.

## Why is this change needed?

When queries built `AffectedTracker`, `ci` remained `false` because `set_ci_check(is_ci())` was not called. In CI environments:
- Tasks configured with `runInCI: only` were incorrectly excluded from affected query results.
- Tasks configured with `runInCI: false` / `skip` were incorrectly included in affected query results.

Fixes #2706.

## How has this been tested?

- Added `can_filter_by_affected_respects_run_in_ci` to `crates/cli/tests/query_tasks_test.rs` covering default, `runInCI: only`, and `runInCI: false` under both CI and local environments.
- Added `respects_run_in_ci` to `crates/cli/tests/query_affected_test.rs` covering CI vs local behavior for `query affected`.
- Ran test suite in Linux Docker (`rust:1.98-bookworm`):
  - `cargo test -p moon_cli --test query_tasks_test --test query_affected_test` (21 passed, 0 failed)
  - `cargo test -p moon_cli --test query_projects_test --test query_changed_files_test` (23 passed, 0 failed)
  - `cargo fmt --check`
  - `cargo clippy -p moon_app -p moon_cli -- -D warnings`


milesj:
@Tyagiquamar To fix the tests, use `is_ci_env` instead of `is_ci`.

PR #2715, opened 2026-09-17
fix: Wait for remote storage to connect before reading or writing the cache

### LINKED ISSUE #2714: [bug] Remote cache lookups silently skip the remote while it is still connecting (2.4.0+)
**Describe the bug**

Since 2.4.0, a task's cache lookup can skip the remote cache entirely and report a miss, even though the entry exists remotely. It happens whenever the lookup runs before moon has finished connecting to the remote. Nothing is logged at any level, including with `MOON_DEBUG_REMOTE=true`.

The cause is how the connection is started and how readability is checked.

* `MoonSession::execute` (`crates/app/src/session.rs`) calls `storage.connect_backends()`. starbase runs `execute` as a background task **alongside** the command, so the pipeline does not wait for it.
* `GrpcRemoteStorage::is_readable` is `false` until `connect()` has set the channel and the capabilities response has set `cache_enabled`. That takes a TCP connection, a TLS handshake and a `GetCapabilities` round trip.
* `Storage::load_manifest` (`crates/cache-storage/src/storage.rs`) skips every backend that is not readable yet, then logs `Cache miss on manifest`. Writes use `is_writable` the same way, so a task that finishes early also skips its upload.

Before #2587, `RemoteService::connect` ran inside the `SyncWorkspace` action. That action always completes before any `RunTask`, so the connection was always ready by the first lookup. The comment there said so: connect "in this action, as it always runs before tasks".

A related issue makes this harder to diagnose. `TaskRunner::is_cached` matches `let Ok(Some(source)) = ...load_manifest(...)`, so a lookup **error** (for example `Unauthenticated`) becomes a miss without any log line.

**Evidence**

We use a gRPC remote cache (BuildBuddy, ~150–190ms per round trip). We run `moon run <project>:<task>` on a single light task in CI with `MOON_LOG=debug`, on 2.5.4. Across 50 runs, measuring from `Connecting to gRPC host` to `Checking for a cache manifest`:

| Lookup started after connecting began | Outcome | Time from `Checking for a cache manifest` to the hit/miss line |
| --- | --- | --- |
| 156–576ms (41 runs) | `Cache miss on manifest` | **0ms** (remote never queried) |
| 397–1086ms (9 runs) | `Cache hit on manifest storage="grpc-remote-cache"` | 150–451ms |

Every entry existed remotely. Re-running the same command from a slower machine hits (`cached from remote`). A real remote miss takes a round trip (~158ms for us), never 0ms.

**Steps to reproduce**

1. Configure a gRPC remote cache whose connection takes a few hundred milliseconds (any non-local host).
2. Populate the cache for a cacheable task with no dependencies in a small workspace.
3. In a fresh checkout, run `MOON_LOG=debug moon run <project>:<task>`.
4. Observe `Checking for a cache manifest` immediately followed by `Cache miss on manifest` with the same timestamp. The task runs instead of hydrating from the remote.

**Expected behavior**

A cache read or write that happens while the remote is still connecting waits for the connection to finish (bounded), then queries the remote. A command that never reaches a cache lookup should not pay for the wait.

**Suggested fix**

We are running this fix in a fork: https://github.com/supertest-ai/moon/compare/v2.5.5...supertest-ai:moon:fix/await-remote-connection

* `Storage` keeps a shared connection state (`NotStarted` / `Connecting` / `Finished`, a `tokio::sync::watch`). `connect_backends` marks it `Connecting`, and a drop guard marks it `Finished`, so a panic or abort cannot leave waiters hanging.
* `spawn_connect_backends()` marks `Connecting` synchronously, then spawns the connection. The session calls it from `analyze`, which is awaited before `execute` starts the command. There is no window where a lookup sees `NotStarted` while a connection is about to begin.
* `load_manifest`, `archive_manifest`, `hydrate_manifest`, `retrieve_blobs` and `store_blobs` wait for an in-progress connection, bounded at 30s. After that they skip unconnected backends with a warning.
* `TaskRunner::is_cached` logs a warning when `load_manifest` returns an error, instead of silently treating it as a miss.
* A storage test starts a lookup before a gated backend has connected. It fails without the change and passes with it.

We ran the same CI restore 6 times on the fork build and got 6 hits. In 4 of the 6, the lookup started before the connection was ready: moon logged `Waiting for storage backends to connect`, then `Cache hit on manifest storage="grpc-remote-cache"` after waiting 300–615ms. On 2.5.4, all 4 would have been misses.

Happy to open a PR with this if the approach works for you.

**Environment**

moon 2.5.4 and 2.5.5 (the code is unchanged on master), Linux x86_64 (musl build) and macOS arm64, gRPC remote cache with header auth.

jkp:
Fixes #2714.

Since 2.4.0, `connect_backends` runs in `MoonSession::execute`, which starbase runs alongside the command. A cache lookup or upload that happens before the connection finishes finds the remote backend unreadable or unwritable, skips it, and reports a miss. Before #2587 the connection ran in `SyncWorkspace`, which always completed before the first task.

## Changes

- **`Storage` tracks connection state** (`NotStarted` / `Connecting` / `Finished`) in a shared `tokio::sync::watch`. `connect_backends` marks it `Connecting`, and a drop guard marks it `Finished`, so a panic or abort can't leave waiters hanging.
- **`Storage::spawn_connect_backends()`** sets `Connecting` synchronously, then spawns the connection. The session calls it from `analyze` (awaited before `execute`) instead of awaiting `connect_backends` in `execute`. The connection still starts as early as before, and a lookup can never see `NotStarted` while a connection is about to begin.
- **Reads and writes wait:** `load_manifest`, `archive_manifest`, `hydrate_manifest`, `retrieve_blobs` and `store_blobs` wait for an in-progress connection, bounded at 30s. After that they skip unconnected backends with a warning. A command that never touches the cache doesn't wait.
- **Lookup errors are logged:** `TaskRunner::is_cached` warns when `load_manifest` returns an error, instead of silently treating it as a miss.

## Testing

- `storage_test.rs` has a new test: `load_manifest_waits_for_a_connection_in_progress` starts a lookup before a gated backend connects. It fails on master and passes with this change. The rest of `moon_cache_storage` passes, and clippy is clean.
- In CI we ran a gRPC remote cache (BuildBuddy) with a release built from this branch. The same `moon run` restore hit the remote 6 times out of 6; on 2.5.4 it hit 9 times out of 50. In 4 of the 6 runs, moon logged `Waiting for storage backends to connect` and then `Cache hit on manifest storage="grpc-remote-cache"`.

I didn't add a changelog entry, since recent fixes leave that to the release. Happy to add one if you'd like.


crates/cache-storage/src/storage.rs
```diff
@@ -68,9 +73,27 @@ impl Default for StorageOptions {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+enum ConnectionState {
+    NotStarted,
+    Connecting,
+    Finished,
```
milesj, line 80: @jkp Do we need a `Connected` state? Because `Finished` doesn't necessarily mean connected if it was dropped.

